PR #1, opened 2025-03-21
Fix install command, avoid assigning lambda to variable

cpsievert:


jcheng5:
Thanks! I did this a slightly different way.

PR #5, opened 2025-04-04
First attempt at genericizing data source

jcheng5:
Do not merge this PR. Already accounted for in #19, #20 via c6a24a7884c7fab47ab7c246af5ca09fe46825a6

Close this PR when #20 is merged.

---

This PR is intended to make querychat extensible beyond in-memory Pandas dataframes, including but not limited to:

1. SQLite
2. PostgreSQL
3. Snowflake
4. Databricks
5. Redshift?
6. Bigquery?

It introduces a protocol called `DataSource`, which so far has two concrete implementations: `DataFrameSource` and `SQLiteSource`.

### TODO

- [ ] Use narwhals instead of Pandas directly
- [ ] Have options to limit the number of rows retrieved, or perhaps retrieve in chunks or a cursor, so big databases won't blow up your Shiny app
- [ ] PostgreSQL
- [ ] Snowflake
- [ ] Databricks

### DIFF
--- a/python-package/examples/app.py
+++ b/python-package/examples/app.py
@@ -4,6 +4,7 @@ from seaborn import load_dataset
 from shiny import App, render, ui
 
 import querychat
+from querychat.datasource import DataFrameSource
 
 titanic = load_dataset("titanic")
 
@@ -14,8 +15,7 @@ with open(Path(__file__).parent / "data_description.md", "r") as f:
 
 # 1. Configure querychat
 querychat_config = querychat.init(
-    titanic,
-    "titanic",
+    DataFrameSource(titanic, "titanic"),
     greeting=greeting,
     data_description=data_desc,
 )
--- a/python-package/pyproject.toml
+++ b/python-package/pyproject.toml
@@ -23,6 +23,11 @@ dependencies = [
     "chevron",
 ]
 
+[project.optional-dependencies]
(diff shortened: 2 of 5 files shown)


jcheng5:
I've decided to focus on SQLAlchemy, which gives us all the features we want for all the databases we care about, as far as I can tell. Now you can do this:

```
pip install SQLAlchemy 
```

```python
from sqlalchemy import create_engine

querychat_config = querychat.init(
    SQLAlchemySource(
        create_engine(f"sqlite:///{Path(__file__).parent / 'titanic.sqlite3'}"),
        "titanic",
    ),
    greeting=greeting,
    data_description=data_desc,
)
```

This may simplify further in the future so you only need to pass the connection string and not call `create_engine` at all, but Claude 3.5 Sonnet feels strongly that this would be surprising for SQLAlchemy users 😄

@blairj09 See https://docs.snowflake.com/en/developer-guide/python-connector/sqlalchemy for doing this with Snowflake.

jcheng5:
We've moved development to `generic-datasource-improvements` since James is presenting off of `generic-datasource` for Snowflake and Databricks conferences these next couple of weeks.

chendaniely:
closing since https://github.com/posit-dev/querychat/pull/20 is merged

PR #10, opened 2025-05-14
fix: No longer need to manually calls `session$ns()` with shinychat

gadenbuie:
With the latest version of shinychat (will be v0.2.0 release), `chat_append_message()` and friends handle namespacing in modules


r-package/DESCRIPTION
```diff
@@ -24,6 +24,8 @@ Imports:
     purrr,
     rlang,
     shiny,
-    shinychat,
+    shinychat (>= 0.1.1.9001),
```
schloerke:
```suggestion
    shinychat (>= 0.2.0),
```

PR #11, opened 2025-05-23
Update Python README to match R

jcheng5:


python-package/README.md
```diff
@@ -46,73 +49,165 @@ app_ui = ui.page_sidebar(
...
+- **Reproducibility:** The SQL query can be easily copied and reused.
+
+Currently, querychat uses DuckDB for its SQL engine. It's extremely fast and has a surprising number of [statistical functions](https://duckdb.org/docs/stable/sql/functions/aggregates.html#statistical-aggregates).
+
+## Customizing querychat
+
+### Provide a greeting (recommended)
+
+When the querychat UI first appears, you will usually want it to greet the user with some basic instructions. By default, these instructions are auto-generated every time a user arrives; this is slow, wasteful, and unpredictable. Instead, you should create a file called `greeting.md`, and when calling `querychat.init`, pass `greeting=Path("greeting.md").read_text()`.
+
```
chendaniely, line 95: You can provide suggestions to the user by using the `<span class="suggestion"> </span>` tag.

For example:

```markdown
* **Filter and sort the data:**
  * <span class="suggestion">Show only survivors</span>
  * <span class="suggestion">Filter to first class passengers under 30</span>
  * <span class="suggestion">Sort by fare from highest to lowest</span>

* **Answer questions about the data:**
  * <span class="suggestion">What was the survival rate by gender?</span>
  * <span class="suggestion">What's the average age of children who survived?</span>
  * <span class="suggestion">How many passengers were traveling alone?</span>
```

This will provide a clickable message in the greeting that will auto populate the chat text box. This gives the user a few ideas to explore on their own. You can see this behavior in our [`querychat template`](https://shiny.posit.co/py/templates/querychat/).

chendaniely: The `<span class="suggestion">` is a good example to point out for the greeting. It really makes the greeting flow better, and we should showcase how to do it. It really polishes up the application and chat.

This was a suggestion of the block of text to put in this part of the README

python-package/README.md
```diff
@@ -46,73 +49,165 @@ app_ui = ui.page_sidebar(
...
+4. (Optional) Any additional instructions you want to use to guide querychat's behavior.
 
-### Configure with your own data
+#### Data description
+
+If you give querychat your dataset and nothing else, it will provide the LLM with the basic schema of your data:
+
+- Column names
+- DuckDB data type (integer, float, boolean, datetime, text)
+- For text columns with less than 10 unique values, we assume they are categorical variables and include the list of values
```
chendaniely: > - For text columns with less than 10 unique values, we assume they are categorical variables and include the list of values. This threshold is configurable.

python-package/README.md
```diff
@@ -46,73 +49,165 @@ app_ui = ui.page_sidebar(
...
+- For integer and float columns, we include the range
+
+And that's all the LLM will know about your data. If the column names are usefully descriptive, it may be able to make a surprising amount of sense out of the data. But if your data frame's columns are `x`, `V1`, `value`, etc., then the model will need to be given more background info--just like a human would.
```
chendaniely: Might be important here to really clarify that the data does not go to the LLM.

> And that's all the LLM will know about your data. The actual data does not get passed into the LLM. We calculate these values before we pass the schema information into the LLM.

python-package/README.md
```diff
@@ -46,73 +49,165 @@ app_ui = ui.page_sidebar(
...
+# Option 2: Use partial
+my_chat_func = partial(chatlas.ChatAnthropic, model="claude-3-7-sonnet-latest")
+
 querychat_config = querychat.init(
-    df=my_dataframe,
+    df=titanic,
+    table_name="titanic",
     create_chat_callback=my_chat_func
 )
 ```
```
chendaniely, line 225: It's really good you have these examples. Shiny prides itself in not having to write a callback, so seeing that term can be a bit scary. I suspect most people will use Option 1.

python-package/README.md
```diff
@@ -46,73 +49,165 @@ app_ui = ui.page_sidebar(
...
-5. Returning filtered data to your Shiny app
+For a complete working example, see the [examples/app.py](examples/app.py) file in the repository. This example includes:
 
-See the `examples/` directory for more complete examples.
\ No newline at end of file
+- Loading a dataset
+- Reading greeting and data description from files
+- Setting up the querychat configuration
+- Creating a Shiny UI with the chat sidebar
+- Displaying the filtered data in the main panel
```
chendaniely: Add at the end.

> If you have Shiny installed, and want to get started right away, you can use our [querrychat template](https://shiny.posit.co/py/templates/querychat/).

PR #28, opened 2025-06-07
feat(r): Generic datasources

npelikan:
This is a WIP but seems to basically work. 

One question -- I retained the basic functionality for local data.frames (that is, df -> querychat -> df), where remote data sources instead return a dbplyr lazy `tbl()`, meant for chaining. Is this too confusing of a behavior split? Should local data.frames also return a `tbl()`, just now connected to duckdb?

A few immediate TODOs:

- [x] add more documentation
- [x] generate some examples
- [x] create shinytests
- [x] validate this works on more than just sqlite


npelikan:
(Changing the target of this to main so it merges cleanly)

pkg-r/R/data_source.R
```diff
@@ -0,0 +1,364 @@
...
+  )
+}
+
+#' Clean up a data source (close connections, etc.)
+#'
+#' @param source A querychat_data_source object
+#' @param ... Additional arguments passed to methods
+#' @return NULL (invisibly)
+#' @export
+cleanup_source <- function(source, ...) {
```
jcheng5, line 233: Should this be `close`?

pkg-r/R/data_source.R
```diff
@@ -0,0 +1,364 @@
...
+  
+  for (col in columns) {
+    col_class <- class(sample_data[[col]])[1]
+    
+    if (col_class %in% c("integer", "numeric", "double", "Date", "POSIXct", "POSIXt")) {
+      numeric_columns <- c(numeric_columns, col)
+      select_parts <- c(
+        select_parts,
+        glue::glue_sql("MIN({`col`}) as {`col`}_min", .con = conn),
+        glue::glue_sql("MAX({`col`}) as {`col`}_max", .con = conn)
```
jcheng5: Let's use more underscores (`__min`) to reduce the possibility of collision.

pkg-r/R/data_source.R
```diff
@@ -0,0 +1,364 @@
...
+        cat_info <- glue::glue("  Categorical values: {values_str}")
+        column_info <- paste(column_info, cat_info, sep = "\n")
+      }
+    } else if (col %in% text_columns) {
+      # For text columns that are not categorical (too many values), still indicate they are categorical
+      # but don't list all the values
+      distinct_count_key <- paste0(col, "_distinct_count")
+      if (distinct_count_key %in% names(column_stats) && !is.na(column_stats[[distinct_count_key]])) {
+        count <- column_stats[[distinct_count_key]]
+        cat_info <- glue::glue("  Categorical values: {count} unique values (exceeds threshold of {categorical_threshold})")
```
jcheng5: Let's not call it "categorical" data in this case, it's just strings. It might or might not be helpful to still say how many unique values. Actually, including the count of unique values would be inaccurate as soon as the data changes. Maybe just leave this line off.

pkg-r/R/querychat.R
```diff
@@ -33,42 +32,31 @@
...
-  }
-
-  force(df)
-  force(system_prompt)
   force(create_chat_func)
-
-  # TODO: Provide nicer looking errors here
+  
+  # Check that data_source is a querychat_data_source object
+  if (!inherits(data_source, "querychat_data_source")) {
```
jcheng5, line 49: If `data_source` is a data frame, let's automatically turn it into the correct `querychat_data_source` for convenience. (Including using the right default table name based on variable)

pkg-r/R/querychat.R
```diff
@@ -156,20 +141,16 @@ querychat_server <- function(id, querychat_config) {
...
 
     current_title <- shiny::reactiveVal(NULL)
     current_query <- shiny::reactiveVal("")
     filtered_df <- shiny::reactive({
-      if (current_query() == "") {
-        df
-      } else {
-        DBI::dbGetQuery(conn, current_query())
-      }
+      querychat::get_lazy_data(data_source, query = dplyr::sql(current_query()))
```
jcheng5: Let's continue to have `querychat$df()` be an eager data frame, and add a new property `querychat$tbl()` for the lazy version.

pkg-r/R/data_source.R
npelikan: Done! Added this for the py package as well.

pkg-r/R/data_source.R
npelikan: Done!

pkg-r/R/querychat.R
npelikan: Done!

pkg-r/R/querychat.R
npelikan: Done!

npelikan:
Alright, I think this is ready to merge @jcheng5

jcheng5:
Shooot, there is one problem left. We're now using `dplyr::tbl(conn, dplyr::sql(query))` to actually perform the SQL queries. The problem is that dplyr assumes that second argument is a table name, not a whole SQL query. So little things that are valid for SQL queries, break; for example, a trailing semicolon, or ending on a SQL comment. You end up seeing pretty confusing error messages with unrecognizable SQL like:

```
SELECT * FROM (SELECT * FROM iris --test) AS `q01` WHERE (0 = 1)
```

I'll ask Hadley what the right way to do this is (if there is one).

npelikan:
Another potential solution to Item 2 could be to implement the defensive programming ourselves, and have it feed sensible errors back to the LLM tools when necessary. I did some reprex-ing in preparation for an issue submission and found that the issue seems to be only around trailing comments on single-line queries and trailing semicolons. It seems that we could solve this with a few simple `gsubs` -- my reprex is below:

```r
library(dplyr)
library(RSQLite)
library(dplyr)
library(DBI)


test_df <- data.frame(
    id = 1:5,
    value = c(10, 20, 30, 40, 50),
    stringsAsFactors = FALSE
  )

temp_db <- tempfile(fileext = ".db")
dbWriteTable(conn, "test_table", test_df, overwrite = TRUE)

# Test with a simple select statement. Works.
simple_select <- tbl(conn, sql("SELECT * FROM test_table WHERE value > 20"))


# Multiline select statement. Works.
multiline_select <- sql("
SELECT *
FROM test_table
WHERE value > 20
")
multiline_select_tbl <- tbl(conn, multiline_select)


# Multiline select statement with trailing inline comment. Works.
multiline_comment <- sql("
SELECT *
FROM test_table
WHERE value > 20 -- this is a filter")
multiline_comment_tbl <- tbl(conn, multiline_comment)


# Trailing semicolon statement. Fails.
trailing_semicolon <- sql("
SELECT *
FROM test_table
WHERE value > 20;")
trailing_semicolon_tbl <- tbl(conn, trailing_semicolon)
# Error in `db_query_fields.DBIConnection()`:
# ! Can't query fields.
# ℹ Using SQL: SELECT * FROM (

# SELECT * FROM test_table WHERE value > 20 ; ) AS `q01` WHERE (0 =
#   1)
# Caused by error:
# ! near ";": syntax error


# single-line select with trailing inline comment
singleline_comment <- sql("SELECT * from test_table WHERE value > 20 --this is a filter")
singleline_comment_tbl <- tbl(conn, singleline_comment)
#  Error in `db_query_fields.DBIConnection()`:
# ! Can't query fields.
# ℹ Using SQL: SELECT * FROM (SELECT * from test_table WHERE value >
#   20 --this is a filter) AS `q02` WHERE (0 = 1)
# Caused by error:
# ! near "WHERE": syntax error


# single-line select statement with midline inline comment. Expected to fail, fails
singleline_midline_comment <- sql("SELECT * from test_table -- breakme WHERE value > 20")
singleline_midline_comment_tbl <- tbl(conn, singleline_midline_comment)
# Error in `db_query_fields.DBIConnection()`:
# ! Can't query fields.
# ℹ Using SQL: SELECT * FROM (SELECT * from test_table -- breakme
#   WHERE value > 20) AS `q03` WHERE (0 = 1)
# Caused by error:
# ! incomplete input

```

gadenbuie:
I think returning a `tbl()` is a great idea and would love to see that realized. Just an idea, but maybe you could wrap the LLM-generated SELECT statement into a view, e.g.

```sql
CREATE OR REPLACE VIEW querychat_results_{table} AS
-- LLM-generated SQL statement
SELECT * FROM {table} WHERE llm_conditions;
```

that you execute with `DBI::dbExecute()`, but then you return a tbl pointing at the view, e.g. `tbl(con, "querychat_reuslts_{table}")`

PR #37, opened 2025-06-18
feat(r+py): Add `querychat_system_prompt(prompt_path=)` support; Export R's `df_to_schema()`

oacar:
I wanted to have more control over default behavior of the system so I added `prompt_path` argument to `querychat_init` and `querychat_system_prompt` to allow user to provide completely custom system prompt.

PR #45, opened 2025-07-01
feat: Rename `prompt_path` to `prompt_template`. Add R support for reading system prompt parameters

schloerke:
Followup to #37

Also:
* Added R support for reading `extra_instructions`, `data_description`, and `prompt_template` if it is an existing file path


schloerke:
cc @oacar  So happy with this final solution! Thank you for the initial push

PR #49, opened 2025-07-29
Classify R Date class variables as duckdb DATE column types in schema

PaulC91:
Hey, thanks for this great pkg.

I noticed the `df_to_schema` function was classifying R Date classes as TEXT columns in the schema passed to the LLM, despite these variables being converted to the duckdb DATE column type in the database. 

This was causing SQL errors when asking the LLM to filter on date columns, so I've added the appropriate check for this in `df_to_schema` and included ranges for this column type in the output. 

I also noticed some incoherence in the readme instructions for including a `data_description` and `extra_instructions`. It suggests to pass the contents of a markdown file via `data_description = readLines("data_description.md")` but this actually causes an error in `read_path_or_string` when there is more than 1 line in the .md file:

```r
> shiny::runApp()
Error in `if (file.exists(x)) ...`:
! the condition has length > 1
``` 

So I've updated the readme to show that only the file path needs to be passed.

I also found it inconsistent that the greeting argument is never passed to `read_path_or_string` and so does need to be passed to `querychat_init` as `greeting = readLines("greeting.md")`. I thought it would be a better UX if `greeting`, `data_description`, and `extra_instructions` could all be passed in the same manner, so I've updated the code and docs to make this change for the `greeting` argument.

Thanks!

### DIFF
--- a/pkg-r/R/prompt.R
+++ b/pkg-r/R/prompt.R
@@ -123,7 +123,7 @@ df_to_schema <- function(
         categories_str <- paste0("'", categories, "'", collapse = ", ")
         info <- c(info, paste0("  Categorical values: ", categories_str))
       }
-    } else if (sql_type %in% c("INTEGER", "FLOAT", "DATETIME")) {
+    } else if (sql_type %in% c("INTEGER", "FLOAT", "DATE", "DATETIME")) {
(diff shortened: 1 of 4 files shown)


PaulC91:
I think recent updates have fixed this issue so I'll close. Thanks.

PR #57, opened 2025-08-07
fix(update_dashboard): Return a confirmation message on successful update

### LINKED ISSUE #56: LLM repeated looping
I'm experiencing an issue where LLMs accessed via `ellmer::chat_ollama` repeat work. Said another way, simple questions are responded to with a correct SQL in the chat window, the dashboard (see [r-sidebot](https://github.com/jcheng5/r-sidebot)) starts to update, but doesn't complete because the LLM starts producing another response...and so on...

```
querychat_handle <- querychat_init(
  tips,
  # This is the greeting that should initially appear in the sidebar when the app
  # loads.
  greeting = readLines(here("greeting.md"), warn = FALSE),
  # create_chat_func = purrr::partial(ellmer::chat_ollama, model = "gpt-oss:20b")
  create_chat_func = purrr::partial(
    ellmer::chat_ollama,
    model = "qwen3:8b"
  )
)
```

More powerful models seem to execute more loops before eventually "tiring" and accepting the last response, which is often the same as all of the previous responses (I've seen up to 60+ using `gpt-oss:20b`).

gadenbuie:
Fixes #56

Prior to this PR, `update_dashboard()` tool doesn't return a value -- or in the R case it returns either `TRUE` or `FALSE` depending on whether the `title` is updated. 

This appears to confuse models, especially local models. This PR updates the tool return value to tell the model the dashboard was updated and to instruct it to use the query tool if needed to see the results.

### DIFF
--- a/pkg-r/R/querychat.R
+++ b/pkg-r/R/querychat.R
@@ -198,6 +198,8 @@ querychat_server <- function(id, querychat_config) {
       if (!is.null(title)) {
         current_title(title)
       }
+
+      "Dashboard updated. Use `query` tool to review results, if needed."
     }
 
     # Perform a SQL query on the data, and return the results as JSON.
@@ -261,7 +263,7 @@ querychat_server <- function(id, querychat_config) {
       # Add user message to the chat history
       shinychat::chat_append(
         "chat",
-        chat$stream_async(input$chat_user_input)
+        chat$stream_async(input$chat_user_input, stream = "content")
       )
     })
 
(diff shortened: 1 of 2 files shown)


pkg-r/R/querychat.R
```diff
@@ -261,7 +263,7 @@ querychat_server <- function(id, querychat_config) {
       # Add user message to the chat history
       shinychat::chat_append(
         "chat",
-        chat$stream_async(input$chat_user_input)
+        chat$stream_async(input$chat_user_input, stream = "content")
```
gadenbuie:
```suggestion
        chat$stream_async(input$chat_user_input)
```

oops, not ready for this yet.

PR #60, opened 2025-08-08
feat(init): Replace `create_chat_func` with `client`

### LINKED ISSUE #58: Replace `create_chat_func` with `client` in `querychat_init()`
I'd like to deprecate the `create_chat_func` in favor of a `client` argument which could be a function, as in `create_chat_func`, but could also be an `ellmer::Chat` object.

This is feasible because we only need to set the system prompt

https://github.com/posit-dev/querychat/blob/e2ced6882092999056d8fd8f17221011cc4f547a/pkg-r/R/querychat.R#L226

which is possible using the `$set_system_prompt()` method.

This does introduce a few new issues:

1. If `client` is a `Chat` object, we'll call `$clone()` to create a new chat.
2. Should we remove turns from the `client`? 
	* I think yes: it maintains current behavior.)
4. Should we remove additional tools from the `client`? 
	* I think maybe not, that could be an interesting way to extend querychat.

gadenbuie:
Fixes #58

## Python

`querychat.init()` now accepts a `client` argument, replacing the previous `create_chat_callback` argument.

The `client` can be:

* a `chatlas.Chat` object,
* a function that returns a `chatlas.Chat` object,
* or a provider-model string, e.g. `"openai/gpt-4.1"`, to be passed to `chatlas.ChatAuto()`.

If `client` is not provided, querychat will use the `QUERYCHAT_CLIENT` environment variable, which should be a provider-model string. If the envvar is not set, querychat uses OpenAI with the default model from `chatlas.ChatOpenAI()`.

## R

`querychat_init()` now accepts a `client`, replacing the previous `create_chat_func` argument.

The `client` can be:

* an `ellmer::Chat` object,
* a function that returns an `ellmer::Chat` object,
* or a provider-model string, e.g. `"openai/gpt-4.1"`, to be passed to `ellmer::chat()`.

If `client` is not provided, querychat will use

* the `querychat.client` R option, which can be any of the above options,
* the `QUERYCHAT_CLIENT` environment variable, which should be a provider-model string,
* or the default model from `ellmer::chat_openai()`.


pkg-py/src/querychat/querychat.py
```diff
@@ -283,7 +356,15 @@ def init(
...
+    client : chatlas.Chat, CreateChatCallback, str, optional
+        A `chatlas.Chat` object, a string to be passed to `chatlas.ChatAuto()`
+        describing the model to use (e.g. `"openai/gpt-4.1"`), or a function
+        that creates a chat client. If not provided, querychat consults the
+        `QUERYCHAT_CLIENT` environment variable, which can be set to a
+        provider-model string. If no option is provided, querychat defaults to
+        using `chatlas.ChatOpenAI(model="gpt-4.1")`.
     create_chat_callback : CreateChatCallback, optional
+        **Deprecated.** Use the `client` argument instead.
         A function that creates a chat object
```
cpsievert:
```suggestion
```

pkg-r/R/querychat_client.R
```diff
@@ -0,0 +1,40 @@
...
+    client <- querychat_client_option()
+  }
+
+  if (is.null(client)) {
+    # Use OpenAI with ellmer's default model
+    return(ellmer::chat_openai())
+  }
+
+  if (rlang::is_function(client)) {
+    client <- client(system_prompt = "")
```
cpsievert: Is the empty string passed here just for back-compat reasons?

pkg-r/R/querychat.R
```diff
@@ -13,8 +13,15 @@
 #' @param data_description A string containing a data description for the chat model. We have found
 #'   that formatting the data description as a markdown bulleted list works best.
 #' @param extra_instructions A string containing extra instructions for the chat model.
-#' @param create_chat_func A function that takes a system prompt and returns a
-#'   chat object. The default uses `ellmer::chat_openai()`.
+#' @param client An `ellmer::Chat` object, a string to be passed to
+#'   [ellmer::chat()] describing the model to use (e.g. `"openai/gpt-4o"`), or
+#'   a function that creates a chat client. If not provided, querychat consults
```
cpsievert: Do we think it's actually useful to also support a function as an argument? FWIW, since there hasn't been a CRAN release, I wouldn't be entirely opposed to not supporting this and dropping `create_chat_func` entirely. If we do, let's just make sure to call it out in the `NEWS.md`

pkg-py/src/querychat/querychat.py
```diff
@@ -517,17 +608,12 @@ async def query(query: str):
 
     chat_ui = ui.Chat("chat")
 
-    # Initialize the chat with the system prompt
-    # This is a placeholder - actual implementation would depend on chatlas
-    chat = create_chat_callback(system_prompt=system_prompt)
+    # Set up the chat object for this session
+    chat = copy.deepcopy(client)
```
cpsievert, line 614:
```suggestion
    chat = copy.deepcopy(client)
    chat.set_turns([])
```

pkg-py/src/querychat/querychat.py
```diff
@@ -233,6 +237,74 @@ def df_to_html(df: IntoFrame, maxrows: int = 5) -> str:
...
+        client = _get_client_from_env()
+
+    if client is None:
+        # Default to OpenAI with using chatlas's default model
+        return chatlas.ChatOpenAI()
+
+    if callable(client) and not isinstance(client, chatlas.Chat):
+        # Backcompat: support the old create_chat_callback style, using an empty
+        # system prompt as a placeholder.
+        client = client(system_prompt="")
```
cpsievert, line 295: Similar to https://github.com/posit-dev/querychat/pull/60/files#r2273954592, I wouldn't be opposed to a breaking change (assuming we document it properly)

pkg-py/src/querychat/querychat.py
```diff
@@ -233,6 +237,74 @@ def df_to_html(df: IntoFrame, maxrows: int = 5) -> str:
...
+    )
+    # We unset chatlas's envvars so we can listen to querychat's envvars instead
+    with temp_env_vars(
+        {
+            "CHATLAS_CHAT_PROVIDER": provider,
+            "CHATLAS_CHAT_MODEL": model,
+            "CHATLAS_CHAT_ARGS": os.environ["QUERYCHAT_CLIENT_ARGS"],
+        },
+    ):
+        return chatlas.ChatAuto(provider="openai")
```
cpsievert, line 261: Just curious, did you have a look into https://github.com/posit-dev/chatlas/issues/90#issuecomment-3168335019 and decide it wasn't worth doing?

pkg-r/R/querychat_client.R
gadenbuie: Yeah, it was relatively common (and I think recommended in the README?) to create a function like `function(system_prompt) { ... }`.

pkg-py/src/querychat/querychat.py
gadenbuie: oops, thanks!

pkg-r/R/querychat.R
gadenbuie: I probably wouldn't have included it if I were writing this from scratch, but I don't have strong feelings that we should take it out either.

pkg-py/src/querychat/querychat.py
gadenbuie: No, I think it's very worth doing. I was just trying to set this up here so that it's not blocked by that work, but in a way that it wouldn't require many changes when we get to posit-dev/chatlas#90.

pkg-r/R/querychat.R
cpsievert: Ok, should probably document what the function signature should look like then?

pkg-r/tests/testthat/test-shiny-app.R
```diff
@@ -9,69 +9,7 @@ test_that("app database example loads without errors", {
   test_app_file <- tempfile(fileext = ".R")
```
cpsievert: Is the intention here that `pkg-r/tests/testthat/apps/basic/app.R` is the new `test_app_file`?

gadenbuie: Oops, I meant to come back to that. Fixed!

cpsievert (approved):

PR #75, opened 2025-09-02
chore(narwhals): Update narwhals import to stable version v1

schloerke:
To avoid breaking changes over time... https://narwhals-dev.github.io/narwhals/backcompat/


gadenbuie (approved):
Nice, thanks!

Issue #70, opened 2025-08-22
Add editable-code box component to be able to edit the SQL query


gadenbuie:
A module function that's attached to a querychat instance that:

1. Shows the SQL code in with highlighting
2. The SQL code is editable
3. Edits update the query being applied
4. Even better: update the query in the tool history (update dashboard tool)
5. Even even better: Update the query and result and regenerate (query tool)

Prior art:

1. I used [Prism Code Editor](https://github.com/FIameCaster/prism-code-editor) for light-weight code editing in [the brand.yml example app](https://github.com/rstudio/bslib/blob/9562108e40a0bffb4a7c8709c2963509435c5c0f/inst/examples-shiny/brand.yml/app.R#L130-L161)
2. That could easily be turned into a module here in querychat that we use for editable code blocks

PR #95, opened 2025-10-15
feat: Add code editor for SQL query

gadenbuie:
Adds `input_code_editor()` with associated `update_code_editor()` that is automatically wired up with `querychat_ui_code()`.

The code editor is a light-weight input using [prism-code-editor](https://github.com/FIameCaster/prism-code-editor).

This is the R package implementation, Python version will be forthcoming.

For #70


gadenbuie:
This feature was moved into bslib as a first-class component: `bslib::input_code_editor()`

Issue #69, opened 2025-08-19
[Python] Add `querychat.app()` function


gadenbuie:
Python version of #59 and #66

cpsievert:
I'm gonna pick this up, since it looks as though Veerle will be creating some querychat+Python content

PR #96, opened 2025-11-11
feat(pkg-py): Expose DataSource's `table_name` attribute

cpsievert:
This will come in handy for #69

Note: R package already has this ability (to get the table name from the data source)

### DIFF
--- a/pkg-py/src/querychat/datasource.py
+++ b/pkg-py/src/querychat/datasource.py
@@ -1,11 +1,10 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, ClassVar, Protocol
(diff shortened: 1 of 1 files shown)


pkg-py/src/querychat/datasource.py
```diff
@@ -87,7 +88,7 @@ def __init__(self, df: IntoFrame, table_name: str):
         """
         self._conn = duckdb.connect(database=":memory:")
         self._df = nw.from_native(df)
-        self._table_name = table_name
+        self.table_name = table_name
```
gadenbuie, line 91: We should consider making this a read-only property; I don't think we want people changing the table name after-the fact

PR #98, opened 2025-11-11
feat(pkg-py): Add programmatic control over the current query/title

cpsievert:
Needed for #69.

Note: R package already has this (albeit via a slightly different API)


gadenbuie (commented):
Looks good, thank you!

PR #99, opened 2025-11-12
feat(pkg-py): Add `querychat.app()`, enable bookmarking by default

cpsievert:
### TODO

 - [ ] Wait for #101 and add as `QueryChat()` method instead?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ maintainers = [
 dependencies = [
     "duckdb",
     "pandas",
-    "shiny",
+    "shiny @ git+https://github.com/posit-dev/py-shiny.git@fix/bookmark-missing-input-error",
     "shinywidgets",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ maintainers = [
 dependencies = [
     "duckdb",
     "pandas",
-    "shiny",
+    "shiny @ git+https://github.com/posit-dev/py-shiny.git@fix/bookmark-missing-input-error",
```
cpsievert, line 24: Once https://github.com/posit-dev/py-shiny/pull/2117 gets merged/released

```suggestion
    "shiny>1.5.0",
```

cpsievert: If we don't want to make release a requirement, should probably make bookmarking an opt-in for now

PR #101, opened 2025-11-12
feat(pkg-py): Add new `QueryChat()` API; hard deprecate old API

### LINKED ISSUE #97: Revisit Python API
Right now the Python API doesn't feel very pythonic (i.e., OOP) -- it feels more like a port of the R API to Python.

I think we should consider something like this:

```python
# Instead of querychat.init()
qc = QueryChat(data)

# Instead of querychat.sidebar()
qc.sidebar()

# Instead of querychat.server()
qc.server()
```

cpsievert:
The PR deprecates essentially the entire current (functional) API in favor of a class-based approach (namely `QueryChat`). The new API looks something like this:

### Express

```python
from seaborn import load_dataset
from shiny.express import render, ui
from querychat.express import QueryChat

titanic = load_dataset("titanic")

qc = QueryChat(titanic, "titanic")
qc.sidebar("chat")

with ui.card(fill=True):
    with ui.card_header():
        @render.text
        def title():
            return qc.title() or "Titanic Dataset"

    @render.data_frame
    def data_table():
        return qc.df()
```

### Core

```python
from seaborn import load_dataset
from shiny import App, render, ui

titanic = load_dataset("titanic")

qc = QueryChat(titanic, "titanic")

app_ui = ui.page_fluid(
    qc.sidebar(),
    ui.output_data_frame("data_table"),
)

def server(input, output, session):
    qc.server()

    @render.data_frame
    def data_table():
        return qc.df()


app = App(app_ui, server)
```

Closes #97 

### TODO

- [ ] Update reference/examples/docs (maybe in follow up PR)?

### DIFF
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changes
 
-* The entire functional API (i.e., `init()`, `sidebar()`, `server()`, etc) has been hard deprecated in favor of a simpler OOP-based API. Namely, the new `QueryChat()` class is now the recommended way to start (instead of `init()`), which has methods to replace old functions (e.g., `.sidebar()`, `.server()`, etc). (#101)
+* The entire functional API (i.e., `init()`, `sidebar()`, `server()`, etc) has been hard deprecated in favor of a simpler OOP-based API. Namely, the new `QueryChat()` class is now the main entry point (instead of `init()`) and has methods to replace old functions (e.g., `.sidebar()`, `.server()`, etc). (#101)
 
 ## [UNRELEASED]
 
--- a/pkg-py/src/querychat/__init__.py
+++ b/pkg-py/src/querychat/__init__.py
@@ -1,4 +1,3 @@
-from . import express
 from ._deprecated import greeting, init, sidebar, system_prompt
 from ._deprecated import mod_server as server
 from ._deprecated import mod_ui as ui
@@ -6,7 +5,6 @@ from ._querychat import QueryChat
 
 __all__ = (
     "QueryChat",
-    "express",
     # TODO(lifecycle): Remove these deprecated functions when we reach v1.0
     "greeting",
     "init",
--- a/pkg-py/src/querychat/_querychat.py
+++ b/pkg-py/src/querychat/_querychat.py
@@ -121,7 +121,6 @@ class QueryChatBase:
                 file=sys.stderr,
             )
 
-        # quality of life improvement to do the Path.read_text() for user or pass along the string
         self.greeting = greeting.read_text() if isinstance(greeting, Path) else greeting
 
         self.system_prompt = get_system_prompt(
--- a/pkg-py/tests/test_querychat.py
+++ b/pkg-py/tests/test_querychat.py
@@ -1,5 +1,3 @@
-"""Tests for the new QueryChat and QueryChatCore API."""
-
 import os
 
 import pandas as pd


cpsievert:
All excellent points, thanks! 

> In Express we run .server() by default; in Core we require that you call .server() in the server function.

I think this would also argue that we have a different class/import for express mode. The difference would be that `.server()` would be internal and runs on `.__init__()`. Or did you have something else in mind?

pkg-py/src/querychat/__init__.py
```diff
@@ -1,14 +1,17 @@
...
+from . import express
+from ._deprecated import greeting, init, sidebar, system_prompt
+from ._deprecated import mod_server as server
+from ._deprecated import mod_ui as ui
+from ._querychat import QueryChat
 
-__all__ = ["greeting", "init", "server", "sidebar", "system_prompt", "ui"]
+__all__ = (
+    "QueryChat",
+    "express",
```
gadenbuie: Should we include `express` in `__all__`? I kind of think it'd be better to require explicitly reaching into `querychat.express`.


```suggestion
    "QueryChat",
```

pkg-py/src/querychat/__init__.py
```diff
@@ -1,14 +1,17 @@
...
+    # Old API (deprecated)
```
gadenbuie: It'd be really helpful to leave a note here that would help us evaluate later when we can remove these (something like date/version where we're deprecating). I like the idea of putting these in `# TODO(lifecyle)` comments so they're easier to find when making lifecycle decisions later.

pkg-py/src/querychat/_querychat_impl.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+class InitResult(TypedDict):
+    data_source: DataSource
+    system_prompt: str
+    greeting: Optional[str]
+    client: chatlas.Chat
+
+
+def init_impl(
```
gadenbuie: Are you using this pattern -- having `QueryChat` methods shell out to `_impl` functions -- to maintain the deprecated functions or because it's a structure you prefer?

If it's primarily about the deprecated functions, I'm inclined to say that we should just have a clean break and do a hard deprecation immediately.

1. The new syntax is so much better; I want to get people into the new flow as quickly as possible.
2. Deprecation warnings are easy to miss; the breaking change forces code upgrades. People can pin to previous versions if they're not ready to migrate yet.
3. We can write up a migration guide now as part of this PR. We tend to rely on deprecation warnings to guide through changes, but I think as a user it's a better experience to be pointed to a page that describes the changes, why they happened and how to migrate.
4. We can keep the deprecated functions as shells that error out and point to that article.

Overall, I think that a clean break from the old API would be better for us in maintaining the code as well as for our users.

pkg-py/src/querychat/__init__.py
cpsievert: By including `express` here, it makes this possible:

```python
import querychat

querychat.express.QueryChat
```

or, somewhat equivalently:

```python
from querychat import *

express.QueryChat
```

It's the same reason why we include `reactive`, `ui`, etc in `shiny/__init__.py`...

pkg-py/src/querychat/_querychat_impl.py
cpsievert: Honestly, I started with soft deprecation since it seemed like the least controversial approach, but you're right that hard deprecating simplifies things significantly enough that I'm up for it 👍. 

See [732d699](https://github.com/posit-dev/querychat/pull/101/commits/732d699caf68456d971024841d3eee495d372f6c)

pkg-py/src/querychat/__init__.py
cpsievert: Done in [732d699](https://github.com/posit-dev/querychat/pull/101/commits/732d699caf68456d971024841d3eee495d372f6c)

pkg-py/src/querychat/__init__.py
gadenbuie: Yeah I mean I don't like either of those and would prefer to discourage them. I think it's much better to have to explicitly import from `express` because you generally don't want both things in the same session. It's different from `reactive` and `ui` in that way.

This seems like the right way to import the express variant:

```python
from querychat.express import Querychat
```

or equivalently

```python
from querychat.express import *
```

and this seems also good to me

```python
import querychat.express

querychat.express.QueryChat
```

pkg-py/src/querychat/_querychat.py
```diff
@@ -0,0 +1,600 @@
...
+
+        if greeting is None:
+            print(
+                "Warning: No greeting provided; the LLM will be invoked at conversation start to generate one. "
+                "For faster startup, lower cost, and determinism, please save a greeting and pass it to init().",
+                "You can also use `querychat.greeting()` to help generate a greeting.",
+                file=sys.stderr,
+            )
+
+        # quality of life improvement to do the Path.read_text() for user or pass along the string
```
gadenbuie:
```suggestion
```

pkg-py/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Changes
+
+* The entire functional API (i.e., `init()`, `sidebar()`, `server()`, etc) has been hard deprecated in favor of a simpler OOP-based API. Namely, the new `QueryChat()` class is now the recommended way to start (instead of `init()`), which has methods to replace old functions (e.g., `.sidebar()`, `.server()`, etc). (#101)
```
gadenbuie: Maybe stronger language around advice; in particular "now the recommended way" is a better fit for the soft deprecation approach

pkg-py/tests/test_querychat_new_api.py
```diff
@@ -0,0 +1,112 @@
+"""Tests for the new QueryChat and QueryChatCore API."""
```
gadenbuie: nit: We can fully commit to this approach, which means it doesn't have to be `querychat_new_api` and can just be The API.

gadenbuie (approved):
This is awesome, very excited to see this land! Just a couple of tiny things before merging

PR #109, opened 2025-11-24
feat!(pkg-r): new `QueryChat` API

cpsievert:
This PR introduces a major API redesign for the R package, replacing the functional API with a simpler R6 class-based approach. These changes are in part motivated by the recent changes made to the Python package (#101 and #108), where the class-based approach is much more idiomatic to Python and Shiny Express. 

###  Summary

  The previous API required users to juggle multiple function calls (`querychat_init()`, `querychat_sidebar()`, `querychat_server()`), explicitly create data sources with `querychat_data_source()`, and route output/input values into the proper locations. The new `QueryChat` R6 class consolidates all this functionality into a single object. 

###  API Comparison

  Before:

```r
  # Create data source
  mtcars_source <- querychat_data_source(mtcars)

  # Configure querychat
  config <- querychat_init(mtcars_source, greeting = "Hello!")

  # Build UI
  ui <- page_sidebar(
    sidebar = querychat_sidebar("chat"),
    DT::DTOutput("data")
  )

  # Server logic
    chat <- querychat_server("chat", config)
    output$data <- renderDT(chat$df())
  }
```

  After:

```r
  # Create QueryChat instance
  qc <- QueryChat$new(mtcars, "mtcars", greeting = "Hello!")

  # Build UI
  ui <- page_sidebar(
    sidebar = qc$sidebar(),
    DT::DTOutput("data")
  )
  
    # Server logic
    qc_valls <- qc$server()
    output$data <- renderDT(qc_vals$df())
  }
```

###  Key Changes

  New QueryChat R6 Class

  - Constructor: `QueryChat$new(data_source, table_name, ...)` accepts data frames or database connections directly
  - UI Methods: `$sidebar()`, `$ui()` create chat interface components
  - Server Method: `$server()` initializes server logic (must be called within Shiny server function)
  - Complete App: `$app()` generates a ready-to-run Shiny app with sensible defaults
  - Greeting Generation: `$generate_greeting()` creates reusable greeting messages
  - Cleanup: `$cleanup()` closes data source resources

###  Hard Deprecations

All previous functional API functions now throw errors with helpful migration messages:

  - `querychat_init()` → `QueryChat$new()`
  - `querychat_app()` → `QueryChat$app()`
  - `querychat_sidebar()` → `QueryChat$sidebar()`
  - `querychat_ui()` → `QueryChat$ui()`
  - `querychat_server()` → `QueryChat$server()`
  - `querychat_greeting()` → `QueryChat$generate_greeting()`
  - `querychat_data_source()` → `QueryChat$new()` (data sources now created internally)

###  Internal Refactoring

  - Renamed `querychat_data_source()` to `as_querychat_data_source()` - now exported as an internal developer extension point only
  - Removed `categorical_threshold` from data source creation (moved to `QueryChat$new()`)
  - Updated all examples and documentation
  - Added new example apps: 01-hello-app/ (minimal usage) and 02-sidebar-app/ (custom UI)

###  Migration Guide

Users upgrading from the old API will receive clear error messages with side-by-side code comparisons showing exactly how to migrate. The new API is more intuitive and requires less code in most cases.

###  Documentation Updates

  - Complete rewrite of README with new API examples
  - Updated all function documentation with proper cross-references
  - Added comprehensive examples for QueryChat class
  - Updated pkgdown configuration

### DIFF
--- a/pkg-r/R/QueryChat.R
+++ b/pkg-r/R/QueryChat.R
@@ -477,7 +477,7 @@ QueryChat <- R6::R6Class(
     #' generate a greeting once and save it for reuse.
     #'
     #' @param echo Whether to print the greeting to the console. Options are
-    #'   `"none"` (default, no output) or `"text"` (print to console).
+    #'   `"none"` (default, no output) or `"output"` (print to console).
     #'
     #' @return The greeting string in Markdown format.
     #'
@@ -487,7 +487,7 @@ QueryChat <- R6::R6Class(
     #' qc <- QueryChat$new(mtcars, "mtcars")
     #'
     #' # Generate a greeting and save it
-    #' greeting <- qc$generate_greeting(echo = "text")
+    #' greeting <- qc$generate_greeting()
     #' writeLines(greeting, "mtcars_greeting.md")
     #'
     #' # Later, use the saved greeting
@@ -640,25 +640,6 @@ querychat <- function(
 #' @return Invisibly returns the chat object after the app stops.
 #'
 #' @export
-#' @examples
-#' \dontrun{
-#' # Quick start - create and run app in one line
-#' querychat_app(mtcars, "mtcars")
-#'
-#' # With options
-#' querychat_app(
-#'   mtcars,
-#'   "mtcars",
-#'   greeting = "Welcome to the mtcars explorer!",
-#'   client = "openai/gpt-4o"
-#' )
-#'
-#' # With database
-#' library(DBI)
-#' conn <- dbConnect(RSQLite::SQLite(), ":memory:")
-#' dbWriteTable(conn, "mtcars", mtcars)
-#' querychat_app(conn, "mtcars")
-#' }
 querychat_app <- function(
   data_source,
   table_name,
--- a/pkg-r/man/QueryChat.Rd
+++ b/pkg-r/man/QueryChat.Rd
@@ -167,7 +167,7 @@ server <- function(input, output, session) {
 qc <- QueryChat$new(mtcars, "mtcars")
 
 # Generate a greeting and save it
-greeting <- qc$generate_greeting(echo = "text")
+greeting <- qc$generate_greeting()
 writeLines(greeting, "mtcars_greeting.md")
 
 # Later, use the saved greeting
@@ -552,7 +552,7 @@ generate a greeting once and save it for reuse.
 \if{html}{\out{<div class="arguments">}}
 \describe{
 \item{\code{echo}}{Whether to print the greeting to the console. Options are
-\code{"none"} (default, no output) or \code{"text"} (print to console).}
+\code{"none"} (default, no output) or \code{"output"} (print to console).}
 }
 \if{html}{\out{</div>}}
 }
@@ -566,7 +566,7 @@ The greeting string in Markdown format.
 qc <- QueryChat$new(mtcars, "mtcars")
 
 # Generate a greeting and save it
-greeting <- qc$generate_greeting(echo = "text")
+greeting <- qc$generate_greeting()
 writeLines(greeting, "mtcars_greeting.md")
 
 # Later, use the saved greeting
--- a/pkg-r/man/querychat-convenience.Rd
+++ b/pkg-r/man/querychat-convenience.Rd
@@ -122,22 +122,4 @@ qc <- querychat(mtcars, "mtcars")
 qc$app()
 
 }
-\dontrun{
-# Quick start - create and run app in one line
-querychat_app(mtcars, "mtcars")
-
-# With options
-querychat_app(
-  mtcars,
-  "mtcars",
-  greeting = "Welcome to the mtcars explorer!",
-  client = "openai/gpt-4o"
-)
-
-# With database
-library(DBI)
-conn <- dbConnect(RSQLite::SQLite(), ":memory:")
-dbWriteTable(conn, "mtcars", mtcars)
-querychat_app(conn, "mtcars")
-}
 }


cpsievert:
@gadenbuie ideally we'd also address the data source plugin API (which is a bit of a mess), but I'm gonna punt on that for now https://github.com/posit-dev/querychat/issues/110

pkg-r/inst/examples-shiny/01-hello-app/app.R
```diff
@@ -0,0 +1,15 @@
...
+app <- qc$app()
+
+# That's it! The app includes:
+# - A sidebar with the chat interface
+# - SQL query display with syntax highlighting
+# - Data table showing filtered results
+# - Reset button to clear queries
+
+# Run with: shiny::runApp()
+app
```
gadenbuie: Printing slash mentioning `runApp()` makes this feel more complicated than it really is

```suggestion
qc$app()

# That's it! The app includes:
# - A sidebar with the chat interface
# - SQL query display with syntax highlighting
# - Data table showing filtered results
# - Reset button to clear queries
```

pkg-r/inst/examples-shiny/02-sidebar-app/app.R
```diff
@@ -0,0 +1,87 @@
...
+
+I can help you explore and analyze the Palmer Penguins dataset. Ask me questions
+about the penguins, and I'll generate SQL queries to get the answers.
+
+Try asking:
+- <span class=\"suggestion\">Show me the first 10 rows of the penguins dataset</span>
+- <span class=\"suggestion\">What's the average bill length by species?</span>
+- <span class=\"suggestion\">Which species has the largest body mass?</span>
+- <span class=\"suggestion\">Create a summary of measurements grouped by species and island</span>
+"
```
gadenbuie: I'm a big fan of R's new raw string literals (from R 4.0, so we can use them freely now)

```suggestion
greeting <- r"(
# Welcome to the Palmer Penguins Explorer! 🐧

I can help you explore and analyze the Palmer Penguins dataset. Ask me questions
about the penguins, and I'll generate SQL queries to get the answers.

Try asking:
- <span class="suggestion">Show me the first 10 rows of the penguins dataset</span>
- <span class="suggestion">What's the average bill length by species?</span>
- <span class="suggestion">Which species has the largest body mass?</span>
- <span class="suggestion">Create a summary of measurements grouped by species and island</span>
)"
```

pkg-r/inst/examples-shiny/02-sidebar-app/app.R
```diff
@@ -0,0 +1,87 @@
...
+
+# Create QueryChat object with custom options
+qc <- QueryChat$new(
+  penguins,
+  "penguins",
+  greeting = greeting,
+  data_description = "The Palmer Penguins dataset contains measurements of bill dimensions, flipper length, body mass, sex, and species (Adelie, Chinstrap, and Gentoo) collected from three islands in the Palmer Archipelago, Antarctica.",
+  extra_instructions = "When showing results, always explain what the data represents and highlight any interesting patterns you observe."
+)
```
gadenbuie, line 35:
```suggestion
qc <- QueryChat$new(
  penguins,
  "penguins",
  greeting = greeting,
  data_description = paste(
    "The Palmer Penguins dataset contains measurements of bill",
    "dimensions, flipper length, body mass, sex, and species",
    "(Adelie, Chinstrap, and Gentoo) collected from three islands in",
    "the Palmer Archipelago, Antarctica."
  ),
  extra_instructions = paste(
    "When showing results, always explain what the data represents",
    "and highlight any interesting patterns you observe."
  )
)

```

pkg-r/R/data_source.R
```diff
@@ -1,28 +1,22 @@
...
 #'   - A character string (e.g., "table_name")
 #'   - Or, for tables contained within catalogs or schemas, a [DBI::Id()] object (e.g., `DBI::Id(schema = "schema_name", table = "table_name")`)
-#' @param categorical_threshold For text columns, the maximum number of unique values to consider as a categorical variable
-#' @param ... Additional arguments passed to specific methods
 #' @return A querychat_data_source object
+#' @keywords internal
 #' @export
-querychat_data_source <- function(x, ...) {
-  UseMethod("querychat_data_source")
+create_data_source <- function(x, table_name = NULL, ...) {
```
gadenbuie: If we're renaming this function, I'd rather `as_querychat_data_source()`. It's verbose, but developer facing, and reflects that it converts from one data source type to another.

If we're going to revamp this flow when we re-evaluate data sources, I'd prefer that we don't rename this function for now

pkg-r/inst/examples-shiny/sqlite/app.R
```diff
@@ -15,8 +15,6 @@ onStop(function() {
 })
 
 conn <- dbConnect(RSQLite::SQLite(), temp_db)
-# The connection will automatically be closed when the app stops, thanks to
-# querychat_init
```
gadenbuie, line 19: Can you comment on the behavior changes you made in this area?

Especially with the new R6 class, I don't think we should be automatically closing this connection -- or at least we _definitely_ shouldn't be doing that when you call the `$app()` method.

We should probably create an example where the querychat object lives entirely within the session; i.e. we connect to the database in each session and use `renderUI()` so that it's entirely session-based. We can also show, in that example, how to clean up the connection.

Otherwise, the more common pattern will be to create the querychat instance in the global environment and use it in both the UI and server. In that use case, it doesn't make sense to clean up the data source except at the session end; it _could_ make sense at app end, but that's something you'd probably want to opt into.

Relatedly, it might be good to have an example that uses pool, or at least to verify that pool would work with querychat.

pkg-r/R/data_source.R
```diff
@@ -87,21 +71,26 @@ querychat_data_source.DBIConnection <- function(
...
+#'
+#' An entrypoint for developers to create custom data source objects for use
+#' with querychat. Most users shouldn't use this function directly; instead,
+#' they call the `$sql()` method on the [QueryChat] object to run queries.
 #'
 #' @param source A querychat_data_source object
 #' @param query SQL query string
 #' @param ... Additional arguments passed to methods
 #' @return Result of the query as a data frame
+#' @keywords internal
```
gadenbuie, line 93: This is a good call for now, but I'm very in favor of a deep overhaul of these functions and of removing them. As dev-facing functions, it's much less likely that anyone has relied on them, unlike the core `querychat_init()` and friends.

pkg-r/R/data_source.R
```diff
@@ -122,10 +111,15 @@ execute_query.dbi_source <- function(source, query, ...) {
...
 #'
+#' An entrypoint for developers to create custom data sources for use with
+#' querychat. Most users shouldn't use this function directly; instead, they
+#' should call the `$sql()` method on the [QueryChat] object to run queries.
+#'
 #' @param source A querychat_data_source object
 #' @param query SQL query string
 #' @param ... Additional arguments passed to methods
 #' @return Result of the query, limited to one row of data.
+#' @keywords internal
```
gadenbuie, line 122: Seems like we could have a `DataSource` class akin to the Python implementation, which could hold these methods/properties (`$execute_query()`, `$test_query()`, `$type`, `$describe()` as a replacement for `create_system_prompt()`, etc.)

pkg-r/R/deprecated.R
```diff
@@ -0,0 +1,158 @@
...
+#' @keywords internal
+NULL
+
+#' @rdname deprecated
+#' @export
+querychat_init <- function(...) {
+  lifecycle::deprecate_stop(
+    when = "0.1.0",
+    what = "querychat_init()",
+    with = "QueryChat$new()",
```
gadenbuie, line 16: This line made me think that I wouldn't be opposed to having a helper function that creates a new `QueryChat` instance. My biggest complaint with `R6` classes is that it is very hard as a user to find docs about the most important step: creating a new instance of the class.

You can't inspect `$new()` to see the arguments and you don't get any help from autocomplete (IIRC). `?QueryChat` often is overfull with context, especially when all you need is a little help to get started.

I'd be okay with `querychat()` as a direct interface to `QueryChat$new()`.

pkg-r/R/deprecated.R
```diff
@@ -0,0 +1,158 @@
...
+  )
+}
+
+#' @rdname deprecated
+#' @export
+querychat_app <- function(...) {
+  lifecycle::deprecate_stop(
+    when = "0.1.0",
+    what = "querychat_app()",
+    with = "QueryChat$app()",
```
gadenbuie: Similar to the above comment, I think it might still be useful to keep `querychat_app()` and have it create the `QueryChat` object and call `$app()` on it.

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+#'   \item{`$df()`}{Get the current filtered data frame (reactive).}
+#'   \item{`$sql(query)`}{Get or set the current SQL query (reactive).}
+#'   \item{`$title(value)`}{Get or set the current title (reactive).}
+#'   \item{`$app()`}{Create a complete Shiny app with sensible defaults.}
+#'   \item{`$generate_greeting(echo)`}{Generate a greeting message using the LLM.}
+#'   \item{`$cleanup()`}{Clean up data source resources.}
+#' }
+#'
+#' @field greeting The greeting message displayed to users.
+#' @field id The module ID for namespacing.
```
gadenbuie: Is this all standard practice? I thought we just document the methods and let pkgdown/R's help create this kind of index?

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+#' }
+QueryChat <- R6::R6Class(
+  "QueryChat",
+  private = list(
+    server_values = NULL,
+    .data_source = NULL,
+    .client = NULL
+  ),
+  public = list(
+    greeting = NULL,
```
gadenbuie, line 74: I've been thinking about this a bit and I'd like to explore having `greeting` be set up as an active binding. We'd have a `private$greeting` that stores the greeting, and then `$greeting()` can be a getter/setter. If we you call `$greeting()` and we don't have one stored yet, we call `generate_greeting()`, store it and then re-use it.

I think this would eliminate the need for `has_greeted` and instead we can restore the greeting.

I guess this brings up the question of how we'd want to handle greetings, for which there are four options.

1. If not provided, always generate a greeting on every new session.
1. If not provided, generate on the first session and then re-use each session.
1. If not provided, don't greet.
1. If provided, always use that greeting.


Should we support all of these? If so, we could have a `generate_greeting = c("once", "always", "never")` argument in `QueryChat$new()`.

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+      rlang::check_dots_empty()
+
+      private$.data_source <- normalize_data_source(data_source, table_name)
+
+      # Validate table name
+      if (!grepl("^[a-zA-Z][a-zA-Z0-9_]*$", table_name)) {
+        rlang::abort(
+          "Table name must begin with a letter and contain only letters, numbers, and underscores"
+        )
+      }
```
gadenbuie, line 161: I think it's worth creating a small `check_sql_table_name()` function.

I also think we should pull in the rlang type checking standalone functions and use them here to validate all of the arguments.

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+        greeting <- paste(readLines(greeting), collapse = "\n")
+      }
+      self$greeting <- greeting
+
+      if (is.null(greeting)) {
+        rlang::warn(c(
+          "No greeting provided; the LLM will be invoked at conversation start to generate one.",
+          "*" = "For faster startup, lower cost, and determinism, please save a greeting and pass it to QueryChat$new().",
+          "i" = "You can generate a greeting with $generate_greeting()."
+        ))
```
gadenbuie, line 175: Let's just go ahead and import and use cli. It's entirely worth it for these messages, and this isn't a package where we need to be extremely concerned about the number of dependencies

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+      private$.client$set_system_prompt(prompt)
+
+      if (auto_cleanup) {
+        # Close the data source when the Shiny app stops (or, if some reason the
+        # querychat_init call is within a specific session, when the session ends)
+        shiny::onStop(function() {
+          message("Closing data source...")
+          self$cleanup()
+        })
+      }
```
gadenbuie: I don't think this should happen in `$init()`, I think this is something we'll want to defer on until later. In particular, I think we'll want `$app()` to make a different decision or we may accidentally break the querychat object.

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+    #' library(querychat)
+    #'
+    #' qc <- QueryChat$new(mtcars, "mtcars")
+    #' app <- qc$app()
+    #'
+    #' # Run the app
+    #' shiny::runApp(app)
+    #'
+    #' # Or return from a script
+    #' app
```
gadenbuie:
```suggestion
    #' qc$app()
    #'
    #' # Or explicitly run the app
    #' shiny::runApp(qc$app())
```

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+
+      table_name <- private$.data_source$table_name
+
+      ui <- function(req) {
+        bslib::page_sidebar(
+          title = shiny::HTML(paste0(
+            "<span>querychat with <code>",
+            table_name,
+            "</code></span>"
+          )),
```
gadenbuie:
```suggestion
          title = shiny::HTML(sprintf(
            "<span>querychat with <code>%s</code></span>",
            table_name
          )),
```

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+    #' \dontrun{
+    #' qc <- QueryChat$new(mtcars, "mtcars")
+    #'
+    #' ui <- fluidPage(
+    #'   qc$ui()
+    #' )
+    #' }
+    ui = function(...) {
+      mod_ui(self$id, ...)
+    },
```
gadenbuie, line 426: Since `...` go to `chat_ui()`, we should probably call `check_dots_named()` here. (Or maybe that happens in `mod_ui()` and I haven't gotten there yet.)

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+    #' qc <- QueryChat$new(mtcars, "mtcars")
+    #'
+    #' server <- function(input, output, session) {
+    #'   qc$server()
+    #'
+    #'   output$data <- renderDataTable(qc$df())
+    #'   output$query <- renderText(qc$sql())
+    #' }
+    #' }
+    server = function(session = shiny::getDefaultReactiveDomain()) {
```
gadenbuie, line 456: Maybe `auto_cleanup` would go here nd in `mod_server()`?

gadenbuie: Also I think `auto_cleanup` could be an enum rather than a bool ([almost always a good idea](https://design.tidyverse.org/boolean-strategies.html)), with options like `c("never", "session_end", "app_end")`

pkg-r/R/QueryChat.R
```diff
@@ -0,0 +1,661 @@
...
+    #'   })
+    #' }
+    #' }
+    title = function(value = NULL) {
+      vals <- private$server_values
+      if (is.null(vals)) {
+        rlang::abort(
+          "Must call $server() before accessing $title(). Make sure to call $server() within your Shiny server function."
+        )
+      }
```
gadenbuie: `$server_value` could be an active binding, too, where the getter wraps this check. Or you can have a `private$get_server_values("title")` method that wraps this up.

pkg-r/R/QueryChat.R
gadenbuie, line 1: Does this file have to be named `QueryChat-class`? Or could it just be `QueryChat.R`?

pkg-r/R/querychat-package.R
```diff
@@ -1,7 +1,61 @@
...
+#'
+#' @section Main Components:
+#' - [QueryChat]: The main R6 class for creating chat interfaces
+#' - [create_data_source()]: (Advanced) Create custom data source objects
+#'
+#' @section Examples:
+#' See the package examples directory for complete working apps:
+#' - `01-hello-app/`: Minimal example using `$app()`
+#' - `02-sidebar-app/`: Custom layout with additional UI elements
+#' - `sqlite/`: Database connection example
```
gadenbuie: I like these sections, nice addition to the package docs! It'd be worth mentioning how to use `shiny::runExample()` to run these examples

pkg-r/R/querychat_module.R
```diff
@@ -0,0 +1,104 @@
+# Main module UI function
+mod_ui <- function(id) {
```
gadenbuie: This needs to take and pass `...` as documented in `$ui()`

pkg-r/R/utils-ellmer.R
```diff
@@ -13,3 +13,47 @@ interpolate_package <- function(path, ..., .envir = parent.frame()) {
 
   ellmer::interpolate_file(path, ..., .envir = .envir)
 }
+
+
+normalize_client <- function(client = NULL) {
```
gadenbuie: nit: I'd prefer `as_querychat_client()` for this function name

pkg-r/NEWS.md
```diff
@@ -1,5 +1,8 @@
 # querychat (development version)
 
+* Nearly the entire functional API (i.e., `querychat_init()`, `querychat_sidebar()`, `querychat_server()`, etc) has been hard deprecated in favor of a simpler OOP-based API. Namely, the new `QueryChat$new()` class is now the main entry point (instead of `querychat_init()`) and has methods to replace old functions (e.g., `$sidebar()`, `$server()`, etc). (#xxx)
+    * In addition, `querychat_data_source()` was renamed to `create_data_source()`, and remains exported for a developer extension point, but users no longer have to explicitly create a data source. (#xxx)
```
gadenbuie:
```suggestion
* Nearly the entire functional API (i.e., `querychat_init()`, `querychat_sidebar()`, `querychat_server()`, etc) has been hard deprecated in favor of a simpler OOP-based API. Namely, the new `QueryChat$new()` class is now the main entry point (instead of `querychat_init()`) and has methods to replace old functions (e.g., `$sidebar()`, `$server()`, etc). (#109)
    * In addition, `querychat_data_source()` was renamed to `create_data_source()`, and remains exported for a developer extension point, but users no longer have to explicitly create a data source. (#109)
```

pkg-r/README.md
```diff
@@ -107,7 +115,7 @@ Currently, querychat uses DuckDB for its SQL engine when working with data frame
 
 ### Provide a greeting (recommended)
 
-When the querychat UI first appears, you will usually want it to greet the user with some basic instructions. By default, these instructions are auto-generated every time a user arrives; this is slow, wasteful, and unpredictable. Instead, you should create a file called `greeting.md`, and when calling `querychat_init`, pass `greeting = readLines("greeting.md")`.
+When the querychat UI first appears, you will usually want it to greet the user with some basic instructions. By default, these instructions are auto-generated every time a user arrives; this is slow, wasteful, and unpredictable. Instead, you should create a file called `greeting.md`, and when creating your `QueryChat` instance, pass `greeting = "greeting.md"` (or use `readLines()` to read the file as a string).
```
gadenbuie: minor nit: this comes off a little judgy, sometimes the variation in greeting is intended and okay.

```suggestion
When the querychat UI first appears, you will usually want it to greet the user with some basic instructions. By default, these instructions are auto-generated every time a user arrives; this is potentially slow, wasteful, and unpredictable. Instead, you should create a file called `greeting.md`, and when creating your `QueryChat` instance, pass `greeting = "greeting.md"` (or use `readLines()` to read the file as a string).
```

pkg-r/README.md
```diff
@@ -183,29 +198,23 @@ performance for 32 automobiles (1973–74 models).
...
-mtcars_source <- querychat_data_source(mtcars, tbl_name = "cars")
-
-# Then initialize with the data source and description
-querychat_config <- querychat_init(
-  data_source = mtcars_source,
-  data_description = readLines("data_description.md")
+qc <- QueryChat$new(
+  mtcars,
+  "mtcars",
+  data_description = "data_description.md"
```
gadenbuie, line 204: This is a nice QoL improvement! (Sae for the handling of `greeting`!)

pkg-r/inst/examples-shiny/sqlite/app.R
gadenbuie: Note: I eventually hit the `auto_close` code in `QueryChat$new()` and left a comment there.

pkg-r/R/data_source.R
cpsievert: Agreed, created an issue for this #110

pkg-r/R/QueryChat.R
cpsievert: Done in fa1c28345ece02fcbf452be0852309b1d6ef9297

pkg-r/R/data_source.R
cpsievert: Yea, I like that, and I also think we should revamp the plugin interface. Since I'm not ready to sign up for that right now, I'd rather we make a small improvement here than nothing at all.

a13172d5fd93f561d060e8ae7af036742548a55c

pkg-r/R/utils-ellmer.R
cpsievert: a13172d5fd93f561d060e8ae7af036742548a55c

pkg-r/R/querychat_module.R
cpsievert: be0ddda90478ba22fcdd427b5fde9bc266c8288d

pkg-r/R/QueryChat.R
cpsievert: Is it possible to support `"app_end"` at this point (where a session is active)?

pkg-r/inst/examples-shiny/sqlite/app.R
cpsievert: Oh, this is a great point, thanks. I hadn't thought through this carefully -- I was just porting what was already going on in `querychat_init()`. See https://github.com/posit-dev/querychat/pull/109#discussion_r2560532727 for some thoughts on this.

pkg-r/R/QueryChat.R
cpsievert: be0ddda90478ba22fcdd427b5fde9bc266c8288d

pkg-r/R/QueryChat.R
cpsievert: I think if we do have some version of `auto_close`, it does actually make the most sense for it to live in `$init()`. That way, we actually have a chance at doing this correctly (only when a session is activate at creation time).

See 1b32ab6af0e3dab59d591a0ad44fc83cbe7c2d5e

pkg-r/R/QueryChat.R
cpsievert: IMO the DRY pattern isn't super worth it in this case.

pkg-r/R/querychat-package.R
cpsievert: c37cba2a356d3b3b7509ed4ab68c847e3e5d5785

pkg-r/R/QueryChat.R
cpsievert: I'd prefer to leave these QoL improvements that are unrelated to the API's design for a follow up PR. I'd like to get this into main ASAP to unblock Veerle, and I'd also like to minimize the amount of logic that we're changing (for git history reasons)

pkg-r/R/QueryChat.R
cpsievert: Same response as https://github.com/posit-dev/querychat/pull/109#discussion_r2561580226

pkg-r/R/deprecated.R
cpsievert: bc157252c2e000b35a9828d7abc3d2d0ec66a35b

pkg-r/R/deprecated.R
cpsievert: bc157252c2e000b35a9828d7abc3d2d0ec66a35b

pkg-r/R/QueryChat.R
cpsievert: Claude suggested this and I'm not sure of another way to get a title inline next to the method name 🤷

pkg-r/R/QueryChat.R
cpsievert: For the sake of making sure the R/Python packages are aligned on this, lets address this in a follow up PR and do both packages at the same time. https://github.com/posit-dev/querychat/issues/113

pkg-r/R/QueryChat.R
```diff
@@ -223,34 +223,63 @@ QueryChat <- R6::R6Class(
...
     #' library(querychat)
     #'
     #' qc <- QueryChat$new(mtcars, "mtcars")
     #' qc$app()
+    #' }
+    #'
+    app = function(..., bookmark_store = "url") {
+      app <- self$app_obj(..., bookmark_store = bookmark_store)
+      tryCatch(shiny::runGadget(app), interrupt = function(cnd) NULL)
+      invisible(private$server_values$chat)
```
gadenbuie: Sorry I missed this in the first review. The advantage of calling `runGadget()` is so that we can return the chat object. Is there another way to expose the session-specific chat clients?

Thinking about that, will our current approach work in apps with simultaneous users?

pkg-r/R/QueryChat.R
gadenbuie: Let's only keep the Methods section, then. The others strike me as non-conventional, but the constructor section isn't formatted correctly (it's inline code rather than code block)

pkg-r/R/QueryChat.R
gadenbuie: idk, maybe we're thinking of different things. I was thinking of the difference between `shiny::onStop()` and `shiny::onSessionEnded()`.

pkg-r/R/QueryChat.R
gadenbuie: For me, it's less about being DRY and more about establishing a pattern that's consistent and easy to follow if we end up adding more of these kinds of accessor functions.

pkg-r/R/QueryChat.R
gadenbuie: I'm conflicted on this; I personally find it helpful that `shinychat::chat_app()` (and `ellmer::live_browser()` by extension) returns the session chat client, but I'm not sure how many other people use this feature and it'd be a lot easier to just return the app object here.

cpsievert: Man I feel silly for not catching this earlier. I think we're gonna have to go back to `$server()` returning a list of reactives

pkg-r/R/QueryChat.R
cpsievert: `shiny::onStop()` is equivalent to `shiny::onSessionEnded()` when used inside the server function....

pkg-r/R/QueryChat.R
cpsievert: See [a5c13ba](https://github.com/posit-dev/querychat/pull/109/commits/a5c13baa41555c4be9229a5ab7b6243ed09c5de9) which does the R equivalent of #116

pkg-r/R/QueryChat.R
cpsievert: Removed everything in [a5c13ba](https://github.com/posit-dev/querychat/pull/109/commits/a5c13baa41555c4be9229a5ab7b6243ed09c5de9). After having a closer look the method section was getting duplicated

pkg-r/R/QueryChat.R
cpsievert: Also, if `stopApp()` is passed a value, `runGadget()` will return it. I also think it's pretty neat that you can get at the most-recent query, so now I'm returning all the session-specific values.

PR #116, opened 2025-11-26
fix(pkg-py): `.server()` now returns reactives, making them truly session-specific

cpsievert:
This PR does the following:

1. Fixes a major issue introduced in #101 by dropping `QueryChat.df()`, `QueryChat.sql()`, etc. in favor of `.server()` now returning those values. I don't think there's a way we can support that API and also prevent one session from effecting another. This doesn't, however, mean we need change the Express API (in that case, the `QueryChatExpress()` instance _is_ session-specific). So, in other words, the Core API now looks like this:

```python
from seaborn import load_dataset
from shiny import App, render, ui

titanic = load_dataset("titanic")

# 1. Provide data source to QueryChat
qc = QueryChat(titanic, "titanic")

app_ui = ui.page_sidebar(
    # 2. Create sidebar chat control
    qc.sidebar(),
    ui.card(
        ui.card_header(ui.output_text("title")),
        ui.output_data_frame("data_table"),
        fill=True,
    ),
    fillable=True
)


def server(input, output, session):
    # 3. Add server logic (to get reactive data frame and title)
    qc_vals = qc.server()

    # 4. Use the filtered/sorted data frame reactively
    @render.data_frame
    def data_table():
        return qc_vals.df()

    @render.text
    def title():
        return qc_vals.title() or "Titanic Dataset"


app = App(app_ui, server)
```

2. As a consequence of 1, it no longer makes sense for `QueryChat` to have a `client` attribute (since you can read the session-specific one via `.server()`). 
3. While here, I decided to add a `system_prompt` property for convenience of inspecting the system prompt (for either class -- if a session-specific one is desired for some reason, it can be obtained via the `client` server value)
4. Added a `types` modules where the return value for `.server()` can be imported, as well as the `DataSource` types. Relatedly, `querychat.datasource` was removed (import from `querychat.types` instead)

### DIFF
--- a/pkg-py/src/querychat/_querychat.py
+++ b/pkg-py/src/querychat/_querychat.py
@@ -386,7 +386,7 @@ class QueryChat(QueryChatBase):
                 return qc_vals.df()
 
             @render.text
-            def title_text():
+            def title():
                 return qc_vals.title() or "My Data"
 
 
@@ -527,7 +527,7 @@ class QueryChatExpress(QueryChatBase):
         session = get_current_session()
         if session is None:
             raise RuntimeError(
-                "Unexpected error: No active Shiny session found."
+                "Unexpected error: No active Shiny session found. "
                 "Is express.QueryChat() being called outside of a Shiny Express app?",
             )
 


pkg-py/src/querychat/_querychat_module.py
```diff
@@ -44,7 +44,36 @@ def mod_ui(**kwargs):
 
 
 @dataclass
-class ModServerResult:
+class ServerValues:
```
gadenbuie, line 47: I'm not sure how much this name matters (since users interact with instances of this class), but I'd rather something like `QueryChatValues`. `ServerValue` seems too generic.

More importantly, I think that returning a dataclass is an excellent compromise and I like this solution as a middle ground to the class instance methods approach we were looking at before.

PR #123, opened 2025-12-01
feat!(pkg-r): new `data_source` plugin API

cpsievert:
# Migrate R Data Source API to R6 Classes

## Overview

This PR migrates the R package's data source API from an S3-based approach to an R6 class-based approach, aligning it with the Python package's design. This is a (developer facing) **breaking change** that provides a cleaner, more maintainable API.

### New R6 Class Hierarchy

**Base Class: `DataSource`**
- Abstract R6 class defining the interface for all data sources
- Methods:
  - `get_db_type()` - Returns database type string
  - `get_schema(categorical_threshold = 20)` - Returns schema information
  - `execute_query(query)` - Executes SQL and returns data frame
  - `test_query(query)` - Tests query by fetching one row
  - `get_data()` - Returns all unfiltered data
  - `cleanup()` - Cleans up resources
- Property:
  - `table_name` - Name of the table

**Concrete Classes:**
- `DataFrameSource` - Wraps data frames using DuckDB
- `DBISource` - Wraps DBI database connections

#### Old API (Removed)
```r
# S3 generic functions
source <- as_querychat_data_source(mtcars, "mtcars")
result <- execute_query(source, "SELECT * FROM mtcars WHERE mpg > 25")
schema <- get_schema(source)
db_type <- get_db_type(source)
cleanup_source(source)
```

#### New API
```r
# Direct R6 instantiation
source <- DataFrameSource$new(mtcars, "mtcars")
result <- source$execute_query("SELECT * FROM mtcars WHERE mpg > 25")
schema <- source$get_schema()
db_type <- source$get_db_type()
source$cleanup()

# Or with DBI connection
source <- DBISource$new(conn, "table_name")
```

### Internal Changes

1. **`data_source.R`** - Complete rewrite with R6 classes
   - Removed all S3 generics
   - Added `DataSource`, `DataFrameSource`, `DBISource` classes
   - Helper functions remain: `get_schema_impl()`, `r_class_to_sql_type()`
   - Renamed `create_system_prompt()` to `get_system_prompt()` (internal)

2. **`QueryChat.R`**
   - Updated `normalize_data_source()` to auto-convert to R6 classes
   - Updated `cleanup()` to call `data_source$cleanup()` method
   - `$data_source` is now an active binding (field), not a method

3. **`querychat_tools.R`**
   - Updated tool implementations to use `$` method syntax
   - Changed from `get_db_type(source)` to `source$get_db_type()`
   - Changed from `execute_query(source, query)` to `source$execute_query(query)`

## Benefits

1. **Consistency**: R and Python packages now share the same design pattern
2. **Better Encapsulation**: Private connection objects are truly private via `private` list
3. **Easier Extensibility**: Users can inherit from `DataSource` for custom implementations instead of defining a handful of (poorly defined) S3 methods.

## Example: Extending DataSource

Users can now easily create custom data sources:

```r
CustomSource <- R6::R6Class(
  "CustomSource",
  inherit = DataSource,
  public = list(
    initialize = function(data, table_name) {
      self$table_name <- table_name
      # Custom initialization
    },

    get_db_type = function() {
      "CustomDB"
    },

    get_schema = function(categorical_threshold = 20) {
      # Custom schema logic
    },

    execute_query = function(query) {
      # Custom query execution
    },

    # ... implement other required methods
  )
)
```

### DIFF
--- a/pkg-r/R/QueryChat.R
+++ b/pkg-r/R/QueryChat.R
@@ -171,7 +171,7 @@ QueryChat <- R6::R6Class(
       }
       self$greeting <- greeting
 
-      prompt <- create_system_prompt(
+      prompt <- assemble_system_prompt(
         private$.data_source,
         data_description = data_description,
         categorical_threshold = categorical_threshold,


pkg-r/R/data_source.R
```diff
@@ -1,282 +1,402 @@
...
-  table_name <- source$table_name
 
+#' Get schema implementation (shared by DataFrameSource and DBISource)
+#'
+#' @param conn DBI connection
+#' @param table_name Table name
+#' @param categorical_threshold Threshold for categorical variables
+#' @return Schema string
+#' @keywords internal
+get_schema_impl <- function(conn, table_name, categorical_threshold = 20) {
```
gadenbuie: Classic Claude... Nothing is lost by removing these "docs"


```suggestion
get_schema_impl <- function(conn, table_name, categorical_threshold = 20) {
```

pkg-r/R/data_source.R
```diff
@@ -1,282 +1,402 @@
...
+
+    #' @description
+    #' Get schema information for the database table
+    #'
+    #' @param categorical_threshold Maximum number of unique values for a text
+    #'   column to be considered categorical (default: 20)
+    #' @return A string describing the schema
+    get_schema = function(categorical_threshold = 20) {
+      get_schema_impl(private$conn, self$table_name, categorical_threshold)
+    },
```
gadenbuie, line 329: I haven't fully tracked where this is used, but I think it's worth caching this (by default at least, maybe with an option to get a fresh schema read). Depending on the data size, this operation could be expensive and it'd be worth caching it if possible

pkg-r/R/data_source.R
```diff
@@ -448,7 +568,11 @@ get_schema.dbi_source <- function(source, categorical_threshold = 20, ...) {
 }
 
 
-# Helper function to map R classes to SQL types
+#' Map R classes to SQL types
+#'
+#' @param r_class R class name
+#' @return SQL type name
+#' @keywords internal
```
gadenbuie:
```suggestion
# Map R classes to SQL types
```

pkg-r/R/data_source.R
```diff
@@ -464,3 +588,51 @@ r_class_to_sql_type <- function(r_class) {
     "TEXT" # default
   )
 }
+
+
+get_system_prompt <- function(
```
gadenbuie: `get_*` isn't the right verb, maybe one of these options?


```suggestion
assemble_system_prompt <- function(
```


```suggestion
prepare_system_prompt <- function(
```


```suggestion
build_system_prompt <- function(
```


```suggestion
render_system_prompt <- function(
```


```suggestion
compose_system_prompt <- function(
```

pkg-r/R/data_source.R
```diff
@@ -464,3 +588,51 @@ r_class_to_sql_type <- function(r_class) {
...
+}
+
+
+read_text <- function(x) {
+  if (file.exists(x)) {
+    x <- readLines(x, warn = FALSE)
+  }
+
+  paste(x, collapse = "\n")
+}
```
gadenbuie, line 627: I think there's something in staticimports that we should pull in for this to make sure we're reading in as UTF-8

pkg-r/tests/testthat/test-querychat-server.R.bak
gadenbuie, line 1: oops, can delete

pkg-r/tests/testthat/test-shiny-app.R
```diff
@@ -35,7 +35,7 @@ test_that("database reactive functionality works correctly", {
   db_conn <- dbConnect(RSQLite::SQLite(), temp_db)
   withr::defer(dbDisconnect(db_conn))
 
-  iris_source <- as_querychat_data_source(db_conn, "iris")
+  iris_source <- DBISource$new(db_conn, "iris")
```
gadenbuie, line 38: This change is fine, but unless I'm missing something this test (about connecting with databases) is out of place in this file (about querychat in Shiny apps)

pkg-r/R/data_source.R
cpsievert: It's currently only called when the system prompt is assembled (on construction of `QueryChat()`) so it seems fine for now

pkg-r/R/data_source.R
cpsievert: Left this for another issue since `readLines()` happens in other places too https://github.com/posit-dev/querychat/issues/134

pkg-r/tests/testthat/test-shiny-app.R
cpsievert: Agreed -- I'm not a fan of it either, but for sake of keeping the diff tractable in this PR, I'm gonna punt on it here

gadenbuie: #135

PR #133, opened 2025-12-03
chore: align and tidy greeting logic

cpsievert:
This better aligns the greeting logic between the 2 packages:

1. Throw warning at greeting creation time (not `QueryChat()` time).
   * R was doing this, but not Python
2. Include auto-generated greeting in the `client`'s turn history (so it can be recovered)
   * Python was doing this, but not R

It also adds a bit of polish to the `greet.qmd` examples

### DIFF
--- a/pkg-r/R/querychat_module.R
+++ b/pkg-r/R/querychat_module.R
@@ -137,5 +137,4 @@ mod_server <- function(
 }
 
 
-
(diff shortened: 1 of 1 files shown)


gadenbuie:
> Include auto-generated greeting in the `client`'s turn history (so it can be recovered)
> * Python was doing this, but not R

I don't think we should do this

PR #149, opened 2025-12-08
chore: `sql()` reactive returns `NULL` or `None` if not set

### LINKED ISSUE #146: Revisit `sql()` returning `""` and `title()` returning `None` when neither are set
It's confusingly inconsistent and I'm not sure there's a good reason for it.

They should probably both return `None`/`NULL`?

gadenbuie:
Fixes #146

This PR updates the `sql()` reactive to be a string or `NULL`/`None`, making `sql()` consistent with `title()`.

The other way to have made these consistent would have been to set them to `""` when unset. That would make checking for emptiness easier, but didn't feel quite right, so I went with the current approach. I largely expect the overall impact to be minimal.

### DIFF
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -9,11 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking Changes
 
-* The `.sql()` method now returns `None` instead of `""` (empty string) when no query has been set, aligning with the behavior of `.title()` for consistency. Most code using the `or` operator or `req()` for falsy checks will continue working without changes. Code that explicitly checks `sql() == ""` should be updated to use falsy checks (`if not sql()`) or explicit null checks (`if sql() is None`). (#146)
-
-### Changes
-
 * The entire functional API (i.e., `init()`, `sidebar()`, `server()`, etc) has been hard deprecated in favor of a simpler OOP-based API. Namely, the new `QueryChat()` class is now the main entry point (instead of `init()`) and has methods to replace old functions (e.g., `.sidebar()`, `.server()`, etc). (#101)
+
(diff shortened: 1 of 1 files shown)


pkg-py/src/querychat/_querychat_module.py
```diff
@@ -113,10 +113,11 @@ def mod_server(
     # Execute query when SQL changes
     @reactive.calc
     def filtered_df():
-        if sql.get() == "":
+        query = sql.get()
+        if not query:
             return data_source.get_data()
         else:
-            return data_source.execute_query(sql.get())
+            return data_source.execute_query(query)
```
gadenbuie, line 120: If we update `.execute_query()` to accept a `None` value, similar to `""`, we could avoid needing to store the result of `sql.get()` in an intermediate variable (to satisfy the type checker).

pkg-py/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Breaking Changes
+
+* The `.sql()` method now returns `None` instead of `""` (empty string) when no query has been set, aligning with the behavior of `.title()` for consistency. Most code using the `or` operator or `req()` for falsy checks will continue working without changes. Code that explicitly checks `sql() == ""` should be updated to use falsy checks (`if not sql()`) or explicit null checks (`if sql() is None`). (#146)
```
cpsievert: Let's put this after the larger breaking change below it.

pkg-py/src/querychat/_querychat_module.py
cpsievert: Yea, seems like that'd be a good idea?

gadenbuie: Nah, that means we'd have to change the return value of `execute_query()` which currently is annotated as `pd.Dataframe`, which then has implications in the base and extender classes. That doesn't seem worth it to me.

cpsievert: Ah ok, the current approach seems fine 👍

PR #152, opened 2025-12-08
feat: `sidebar()` is now fillable by default

cpsievert:
Before this change: the chat input appears immediately below the greeting. 

After this change: the chat input is anchored at the bottom of the viewport.

```r
querychat::querychat_app(mtcars)
```

Screenshot 2025-12-08 at 5 21 36 PM

### DIFF
--- a/pkg-r/man/QueryChat.Rd
+++ b/pkg-r/man/QueryChat.Rd
@@ -414,7 +414,7 @@ Create a sidebar containing the querychat UI.
 This method generates a \code{\link[bslib:sidebar]{bslib::sidebar()}} component containing the chat
 interface, suitable for use with \code{\link[bslib:page_sidebar]{bslib::page_sidebar()}} or similar layouts.
 \subsection{Usage}{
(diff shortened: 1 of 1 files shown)


gadenbuie (approved):
Thanks!

pkg-r/R/QueryChat.R
```diff
@@ -396,10 +397,11 @@ QueryChat <- R6::R6Class(
     #'   # Main content here
     #' )
     #' }
-    sidebar = function(width = 400, height = "100%", ...) {
+    sidebar = function(width = 400, height = "100%", fillable = TRUE, ...) {
       bslib::sidebar(
         width = width,
         height = height,
+        fillable = fillable,
```
gadenbuie, line 404: Does this mean we need to release bslib in order to release querychat? Or can we get away with releasing querychat first?

cpsievert: Oh, good point -- I don't think we'd have to make it a strict requirement since it's a progressive enhancement, but I also wouldn't be opposed to getting another release out

gadenbuie: Cool -- I want to get bslib out by end of year but I'd rather push querychat out first if we can.

PR #158, opened 2025-12-09
fix(pkg-r): `DBI::SQL()` errors given `NULL`

cpsievert:
Introduced by #149. I'm pretty sure the `DBI::SQL()` isn't needed in the first place, so we can just drop it.

Here's a reprex:

```r
library(shiny)
library(bslib)
library(querychat)
library(palmerpenguins)

qc <- QueryChat$new(penguins)

ui <- page_sidebar(
  sidebar = qc$sidebar(),
  dataTableOutput("table")
)

  qc_vals <- qc$server()
  
  output$table <- renderDataTable({
    qc_vals$df()
  })
}

shinyApp(ui, server)
```

```
Warning: Error in initialize: cannot use object of class “NULL” in new():  class “SQL” does not extend that class
```


cpsievert:
@gadenbuie merging to un-break `main`, but please comment if you'd like

PR #162, opened 2025-12-10
feat(pkg-r): website improvements

### LINKED ISSUE #128: (R) Update website
Borrowing ideas from #119

cpsievert:
Closes #128 

Some of the Python changes here are a follow up to #119

### DIFF
--- a/pkg-py/docs/build.qmd
+++ b/pkg-py/docs/build.qmd
@@ -26,7 +26,7 @@ If you have an existing app with a reactive data frame that flows through multip
 Integrating querychat into a Shiny app requires just three steps:
 
 1. Initialize a `querychat()` instance with your data
-2. Add the querychat UI component (either `.sidebar()` or `.ui()`)
+2. Add the UI component (either `.sidebar()` or `.ui()`)
 3. Use reactive values like `.df()`, `.sql()`, and `.title()` to build outputs that respond to user queries
 
 Here's a starter template demonstrating these steps:
@@ -279,8 +279,8 @@ This is equivalent to the user asking the LLM to "reset" or "show all data".
 
 Currently, you have two options for exploring multiple tables in querychat:
 
-1. Join the tables into a single table before passing to QueryChat
-2. Use multiple QueryChat instances in the same app
+1. Join the tables into a single table before passing to querychat
+2. Use multiple querychat instances in the same app
 
 The first option makes it possible to chat with multiple tables inside a single chat interface, whereas the second option requires a separate chat interface for each table.
 
@@ -290,7 +290,7 @@ The first option makes it possible to chat with multiple tables inside a single 
 We plan to support multiple filtered tables in a future release -- if you're interested in this feature, please upvote [the relevant issue](https://github.com/posit-dev/querychat/issues/6)
 :::
 
-Here's an example of the second approach, using two separate QueryChat instances to explore both the `titanic` and `penguins` datasets within the same app:
+Here's an example of the second approach, using two separate querychat instances to explore both the `titanic` and `penguins` datasets within the same app:
 
 ```{.python filename="multiple-datasets.py" code-fold="true" code-summary="Show app code"}
 {{< include _examples/multiple-datasets.py >}}


docs/index.html
```diff
@@ -222,9 +222,9 @@
                 </a>
             </div>
             <h1 class="package-title">querychat</h1>
-            <p class="package-subtitle">Chat with your data in any language</p>
+            <p class="package-subtitle">Explore data using natural language queries</p>
```
gadenbuie: Ooh, I actually like the original better here. "chat with your data" is fun, sounds inviting and is a pretty good description of what querychat does. "Explore" sounds a little too broad for what querychat actually handles. 

And "in any language" is a great way to cover both the R and Python language coverage as well as human languages, since LLMs are generally multi-lingual.

I'm not a big fan of the phrase "natural language queries" unless we're intentionally getting Very Technical. It works in the description as a counterpoint to the more vague tagline, but as a tagline it's kind of boring.

pkg-py/docs/build.qmd
```diff
@@ -346,9 +346,25 @@ def _():
...
+
+1. Join the tables into a single table before passing to QueryChat
+2. Use multiple QueryChat instances in the same app
+
+The first option makes it possible to chat with multiple tables inside a single chat interface, whereas the second option requires a separate chat interface for each table.
+
+::: {.callout-note}
+### Multiple filtered tables
+
+We do intend on supporting multiple filtered tables in a future release -- if you're interested in this feature, please up vote [the relevant issue](https://github.com/posit-dev/querychat/issues/6)
```
gadenbuie:
```suggestion
We plan to support multiple filtered tables in a future release -- if you're interested in this feature, please up vote [the relevant issue](https://github.com/posit-dev/querychat/issues/6)
```

pkg-py/docs/build.qmd
```diff
@@ -346,9 +346,25 @@ def _():
...
+:::
+
+Here's an example of the second approach, using two separate `QueryChat` instances to explore both the `titanic` and `penguins` datasets within the same app:
+
+<details>
+<summary> <code>app.py </code> </summary>
```
gadenbuie: Two quick things because we're using Quarto here:

1. We should put these little apps in their own source files and include them using `{{< include app-two-chats.py >}}`. That makes it much easier to run and test the apps outside of the Quarto document.
2. Instead of using `<details><summary>` elements, we can lean on Quarto's code folding. It's functionally similar but visually I think it will look better. If not, we could use callouts. The bare details elements don't look so great.

pkg-py/docs/context.qmd
```diff
@@ -2,13 +2,13 @@
 title: Provide context
 ---
 
-To improve the LLM's ability to accurately translate natural language queries into SQL, it often helps to provide relevant metadata. Querychat automatically provides things like column names and data types to the LLM, but you can enhance this further with additional context like [data descriptions](#data-description). You can also provide [custom instructions](#extra-instructions) to add additional behaviors and even supply a fully [custom prompt template](#custom-template), if desired.
+To improve the LLM's ability to accurately translate natural language queries into SQL, it often helps to provide relevant metadata. QueryChat automatically provides things like column names and data types to the LLM, but you can enhance this further with additional context like [data descriptions](#data-description). You can also provide [custom instructions](#extra-instructions) to add additional behaviors and even supply a fully [custom prompt template](#custom-template), if desired.
```
gadenbuie: There presentation of ideas here seems to have some overlapping qualities. I think it'd be clearer if it were a little more linear:

1. QueryChat automatically gathers some information about your table -- column names and types, numerical ranges and categorical value examples.
2. LLMs don't see the actual data, and they don't need to in order to write SQL queries for you
3. You can get even better results by customizing the system prompt and there are three ways you can do this:
    1. Add more description about the data
    2. Add more instructions
    3. Go fully custom with a template (use this if you want to be certain that the model cannot see any literal values from your data)

gadenbuie: The definition of _system prompt_ in the paragraph below could be a footnote.

pkg-py/docs/context.qmd
```diff
@@ -2,13 +2,13 @@
...
 
 All of this information is provided to the LLM as part of the **system prompt** -- a string of text containing instructions and context for the LLM to consider when responding to user queries.
 
 ## Default prompt
 
-For full visibility into the full system prompt that Querychat generates for the LLM, see the `system_prompt` property. This is useful for debugging and understanding exactly what context the LLM is using:
+For full visibility into the full system prompt that QueryChat generates for the LLM, see the `system_prompt` property. This is useful for debugging and understanding exactly what context the LLM is using:
```
gadenbuie: It feels weird to me that we're calling it `QueryChat` but the package name is `querychat`. `pip install QueryChat` works but `install.packages("QueryChat")` won't.

I like the effort for consistency! I also generally like the casing for accessibility reasons, but I'm not sure I really want to maintain separate formatting for the display name and package name, especially when it's `querychat` in both languages

pkg-py/docs/index.qmd
```diff
@@ -22,8 +22,7 @@ Explore data using natural language queries
 <!-- badges end -->
 </div>
 
-
-Querychat makes it easy to explore data with natural language through the power of [Shiny](https://shiny.posit.co/py) and large language models (LLMs). Start chatting with your data in just one line of code. Or, with a few more lines, design your own rich user experience around data exploration and analysis through natural language.
+QueryChat facilitates safe and reliable natural language exploration of tabular data, powered by SQL and large language models (LLMs). For analysts, it offers an intuitive web application where they can quickly ask questions of their data and receive verifiable data-driven answers. For software developers, QueryChat provides a comprehensive Python API to access core functionality -- including chat UI, generated SQL statements, resulting data, and more. This capability enables the seamless integration of natural language querying into bespoke data applications.
```
gadenbuie: "For analysts" is too limiting; you can give a querychat app to regular people too. I think the idea here was "For users" or "For your app users".

> For software developers, QueryChat provides a comprehensive Python API to access core functionality.

I'm assuming Claude helped write this because it doesn't say much. Of course a Python package will give you access to its core functionality.

---

I like the direction of this paragraph: users on the one hand, developers on the other. But I think it needs a rewrite. It's okay for this paragraph to target developers primarily. Just switching to second person, where "you" means "software developers", would help. It's a bit "at a distance" right now

pkg-py/docs/index.qmd
```diff
@@ -45,6 +44,7 @@ from querychat.data import titanic
 
 qc = QueryChat(titanic(), "titanic", client="openai/gpt-4.1")
 app = qc.app()
+# app.run()
 ```
 
 With the above code saved to `titanic-app.py` and an API key set[^api-key], you can [run the app](https://shiny.posit.co/py/get-started/create-run.html#run-your-shiny-application) from a terminal (or [VSCode](https://marketplace.visualstudio.com/items?itemName=Posit.shiny)):
```
gadenbuie: Is it VSCode or VS Code? Turns out there's [a right answer](https://code.visualstudio.com/brand): `VS Code`.

Also, I think we should mention Positron first everywhere it comes up and add `(or VS Code)` if we need to mention VS Code.

pkg-py/docs/index.qmd
gadenbuie: I kind of think we should show the `QUERYCHAT_CLIENT` envvar here too. It's not strictly necessary but it makes it very clear from the start that you can use querychat with any model/provider supported by chatlas. It'd be nice to have that front and center here.

gadenbuie: Yes, you set `client` in the example above, but actually I think you could drop that and just use `QueryChat(titanic(), "titanic")`. That makes getting started feel a lot easier

pkg-py/docs/index.qmd
gadenbuie: The transition here – "With the above code saved to..." – makes it feel like saving the file is a necessary step. If you add a little transition phrase, like "You can run that code in a Python console and then call `app.run()` to jump into a chat. Or you can save the code to a file and ..." you can separate the two parts

pkg-py/docs/context.qmd
cpsievert: Are you suggesting we use querychat (for display name) everywhere (even at the start of a sentence/title)? And `querychat` if referencing a specific package?

pkg-py/docs/index.qmd
```diff
@@ -68,20 +68,25 @@ Suppose we pick a suggestion like "Show me passengers who survived". Since this
 
 ![](/images/quickstart-filter.png){fig-alt="Screenshot of the querychat's app with the titanic dataset filtered to passengers who survived." class="lightbox shadow rounded mb-3"}
 
-Querychat can also handle more general questions about the data that require calculations and aggregations. For example, we can ask "What is the average age of passengers who survived?". In this case, querychat will generate/execute the SQL query to perform the relevant calculation, and return the result in the chat:
+QueryChat can also handle more general questions about the data that require calculations and aggregations. For example, we can ask "What is the average age of passengers who survived?". In this case, QueryChat will generate/execute the SQL query to perform the relevant calculation, and return the result in the chat:
```
gadenbuie:
```suggestion
QueryChat can also handle more general questions about the data that require calculations and aggregations. For example, we can ask "What is the average age of passengers who survived?". QueryChat will generate/execute the SQL query to perform the relevant calculation, and return the result in the chat:
```

"In this case" isn't needed here.

But generally reading this I realized that there's a blurring happening between QueryChat and the LLM, i.e. QueryChat won't generate SQL, the LLM does, but QueryChat does execute the SQL. Since I'm reading through this in a diff, I'm just going to bring it up; maybe it's okay at this place in the docs.

pkg-py/docs/index.qmd
```diff
@@ -68,20 +68,25 @@ Suppose we pick a suggestion like "Show me passengers who survived". Since this
...
 
 ![](/images/quickstart-summary.png){fig-alt="Screenshot of the querychat's app with a summary statistic inlined in the chat." class="lightbox shadow rounded mb-3"}
 
-As you'll learn later in [Build an app](build.qmd), you can also access the SQL query and filtered/sorted data frame programmatically for use elsewhere in your app. This makes it rather seemless to have natural language interaction with your data alongside other visualizations and analyses.
+## Custom apps
+
+QueryChat is designed to be highly extensible -- it provides programmatic access to the chat interface, the filtered/sorted data frame, SQL queries, and more.
+This makes it easy to build custom web apps that (safely) leverage natural language interaction with your data.
```
gadenbuie: kind of a weird moment to bring up "safely". I think you can just drop it


```suggestion
This makes it easy to build custom web apps that leverage natural language interaction with your data.
```

pkg-py/docs/index.qmd
```diff
@@ -68,20 +68,25 @@ Suppose we pick a suggestion like "Show me passengers who survived". Since this
...
 
-Before we build though, let's take a moment to better understand how querychat works under the hood, and whether it's right for you.
+![](/images/airbnb.png){fig-alt="A custom app for exploring Airbnb listings, powered by QueryChat." class="lightbox shadow rounded mb-3"}
 
+To learn more, see [Build an app](build.qmd) for a step-by-step guide.
 
 ## How it works
 
-Querychat leverages LLMs incredible capability to translate natural language into SQL queries. Frontier models are shockingly good at this task, but even the best models still need to know the overall data structure to perform well. For this reason, querychat supplies a [system prompt](context.qmd) with the schema of the data (i.e., column names, types, ranges, etc), but never the raw data itself.
+QueryChat leverages LLMs incredible capability to translate natural language into SQL queries. Frontier models are shockingly good at this task, but even the best models still need to know the overall data structure to perform well. For this reason, QueryChat supplies a [system prompt](context.qmd) with the schema of the data (i.e., column names, types, ranges, etc), but never the raw data itself.
```
gadenbuie:
```suggestion
QueryChat leverages the incredible capabilities of large language models to translate natural language into SQL queries.
Models of all sizes, from small models you can run locally on your laptop to large frontier models by major AI providers, are all surprisingly good at this task.

That said, even the best models still need to know the overall data structure to perform well.
For this reason, QueryChat includes a description of the data's schema -- column names, types, ranges, categorical values -- in the LLM's [system prompt](context.qmd).
Note that QueryChat **does not** send the raw data itself to the model! 
Instead, it shares just enough of a description of the data that the model can reliably generate SQL queries.
```

pkg-py/docs/index.qmd
```diff
@@ -68,20 +68,25 @@ Suppose we pick a suggestion like "Show me passengers who survived". Since this
...
 
-When the LLM generates a SQL query, querychat executes it against a SQL database (DuckDB[^duckdb] by default) to get results in a **safe**, **reliable**, and **verifiable** manner. In short, this execution is **safe** since only `SELECT` statements are allowed, **reliable** since the database engine handles all calculations, and **verifiable** since the user can always see the SQL query that was run. This makes querychat a trustworthy tool for data exploration, as every action taken by the LLM is transparent and independently reproducible.
+When the LLM generates a SQL query, QueryChat executes it against a SQL database (DuckDB[^duckdb] by default) to get results in a **safe**, **reliable**, and **verifiable** manner. In short, this execution is **safe** since only `SELECT` statements are allowed, **reliable** since the database engine handles all calculations, and **verifiable** since the user can always see the SQL query that was run. This makes QueryChat a trustworthy tool for data exploration, as every action taken by the LLM is transparent and independently reproducible.
```
gadenbuie:
```suggestion
When the LLM generates a query, QueryChat executes it against a SQL database (DuckDB[^duckdb] by default) to get results in a **safe**, **reliable**, and **verifiable** manner. 
In short, this execution is **safe** since only `SELECT` statements are allowed, **reliable** since the database engine handles all calculations, and **verifiable** since the user can always see the SQL query that was run.
This makes QueryChat a trustworthy tool for data exploration, as every action taken by the LLM is transparent and independently reproducible.
```

> this execution is **safe** since only `SELECT` statements are allowed

This bit gives me pause because it's imposed by prompting, but this sounds like we'd actually take steps to disallow statements that don't rely on `SELECT`

pkg-py/docs/tools.qmd
```diff
@@ -4,16 +4,16 @@ title: Tools
 
 QueryChat combines [tool calling](https://posit-dev.github.io/chatlas/get-started/tools.html) with [reactivity](https://shiny.posit.co/py/docs/reactive-foundations.html) to not only execute SQL, but also reactively update dependent data views. Understanding how these tools work will help you better understand what QueryChat is capable of and how to customize/extend to its behavior. 
 
-One important thing to understand generally about Querychat's tools is they are Python functions, and that execution happens on _your machine_, not on the LLM provider's side. In other words, the SQL queries generated by the LLM are executed locally in the Shiny app process, and only the results (if any) are sent back to the LLM.
+One important thing to understand generally about QueryChat's tools is they are Python functions, and that execution happens on _your machine_, not on the LLM provider's side. In other words, the SQL queries generated by the LLM are executed locally in the Shiny app process, and only the results (if any) are sent back to the LLM.
```
gadenbuie:
```suggestion
One important thing to understand generally about QueryChat's tools is they are Python functions, and that execution happens on _your machine_, not on the LLM provider's side. In other words, the SQL queries generated by the LLM are executed locally in the Shiny app process.
```

It's clarified below but this phrasing sets up "and sends the results back to the LLM" as the default behavior, but really it's only in the case of the `query` tool

pkg-py/docs/tools.qmd
```diff
@@ -4,16 +4,16 @@ title: Tools
...
 
-Querychat provides the LLM access to three tools, serving two primary purposes:
+QueryChat provides the LLM access to three tools, serving two primary purposes:
```
gadenbuie:
```suggestion
QueryChat provides the LLM access to two tool groups:
```

pkg-py/docs/tools.qmd
```diff
@@ -4,16 +4,16 @@ title: Tools
...
 
 ## Data updating
 
-When a user asks to "Show me..." or "Filter to..." or "Sort by...", the LLM requests a call to the `update_dashboard` tool with an appropriate SQL query as input. An important constraint is that the query must return all original schema columns (typically using `SELECT *`). When called, Querychat will both set a reactive value holding [the current SQL query](build.qmd#sql-query) and execute the query to get the result.
+When a user asks to "Show me..." or "Filter to..." or "Sort by...", the LLM requests a call to the `update_dashboard` tool with an appropriate SQL query as input. An important constraint is that the query must return all original schema columns (typically using `SELECT *`). When called, QueryChat will both set a reactive value holding [the current SQL query](build.qmd#sql-query) and execute the query to get the result.
 
 The result of query then used to set a reactive value holding the [filtered/sorted data frame](build.qmd#filtered-data). Thanks to reactivity, this will automatically update any views depending on this data frame, such as the data table displayed in the UI.
```
gadenbuie: The last sentence of para 1 and first sentence of para 2 should probably be together. 

As a Shiny app dev I'd have a hard time relating to the phrase "set a reactive value holding the X". I think it'd be fine to say QueryChat executes the query and makes the results and the source query available as reactive values that you can use in your app. Whenever the values update, so does everything else in your app.

Maybe we could use a diagram around these parts, too?

pkg-py/docs/context.qmd
gadenbuie: oh for a concrete recommendation, I'd say to use querychat everywhere (i.e. tidyverse formatting, where package names don't get special treatment). So yeah, querychat in starts of sentences and titles. 

Example:

> querychat can be installed with `pip install querychat`

gadenbuie (commented):
Good stuff! It's definitely a big step forward for the docs. I've been submitting feedback as I worked through things. I made it through the stuff that shows up in the Python diffs; I'll pick up with the R things in a bit, although I suspect there's some overlap and that some of the Python comments will directly translate to the R docs

pkg-py/docs/index.qmd
cpsievert: It's true that we currently don't do anything, but I'm thinking we've leave this as-is for now, and possibly do more to ensure this?

cpsievert: I walked back this wording (and revisited this section more generally) in [e8cc12e](https://github.com/posit-dev/querychat/pull/162/commits/e8cc12ec0e18dafe30e2c188ac8339f0664bd1d7)

pkg-r/vignettes/build.Rmd
```diff
@@ -0,0 +1,443 @@
...
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  eval = FALSE
+)
+```
+
+While `querychat_app()` provides a quick way to start exploring data, building bespoke Shiny apps with QueryChat unlocks the full power of integrating natural language data exploration with custom visualizations, layouts, and interactivity. This guide shows you how to integrate QueryChat into your own Shiny applications and leverage its reactive data outputs to create rich, interactive dashboards.
```
gadenbuie: I think it's a good idea to start with the simple template, but this article in general assumes you're starting from scratch to build a Shiny app that wraps querychat.

I think it'd be useful to talk about what kinds of apps make good querychat apps up front, which would also help people who have an existing app they want to bring querychat into.

This was the approach that I took in the Programming with LLMs workshop, some of my slides might help: https://posit-conf-2025.github.io/llm/slides/slides-10.html#/querychat

The general idea is to acknowledge that apps that have a single data source plus a bunch of filters that combine to create a reactive data frame that is used in a lot of different places is probably the best use case for a querychat powered Shiny app. (That's the idea with the first two diagrams, at least.)

pkg-r/vignettes/context.Rmd
```diff
@@ -0,0 +1,120 @@
...
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  eval = FALSE
+)
+```
+
+querychat automatically gathers information about your table to help the LLM write accurate SQL queries. This includes column names and types, numerical ranges, and categorical value examples. (All of this information is provided to the LLM as part of the **system prompt** -- a string of text containing instructions and context for the LLM to consider when responding to user queries.)
+
+Importantly, the LLM never sees the actual data itself -- it doesn't need to in order to write SQL queries for you. It only needs to understand the structure and schema of your data.
```
gadenbuie: Let's flip this and say something like "we are **not** sending your raw data to the LLM and asking it to do complicated math".

There's nuance in the "never sees the actual data itself" that I think we're better off avoiding and we can explain the fundamentals without having to contradict ourselves.

gadenbuie: Could also be good to point to an article here, something about how LLMs are bad at math, like Wes' recent post.

* [Can LLMs give us AGI if they are bad at arithmetic? – Wes McKinney](https://wesmckinney.com/blog/llms-arithmetic/)
* [Everything we know about LLMs doing Arithmetic](https://loeber.substack.com/p/21-everything-we-know-about-llms)

pkg-r/vignettes/context.Rmd
```diff
@@ -0,0 +1,120 @@
...
+3. Use a fully [custom prompt template](#custom-template) if you want complete control (useful if you want to be certain the model cannot see any literal values from your data)
+
+```{r}
+library(querychat)
+library(palmerpenguins)
+```
+
+## Default prompt
+
+For full visibility into the full system prompt that querychat generates for the LLM, see the `system_prompt` field. This is useful for debugging and understanding exactly what context the LLM is using:
```
gadenbuie:
```suggestion
For full visibility into the system prompt that querychat generates for the LLM, you can inspect the `system_prompt` field. This is useful for debugging and understanding exactly what context the LLM is using:
```

pkg-r/vignettes/context.Rmd
```diff
@@ -0,0 +1,120 @@
...
+  data_description = "data_description.md"
+)
+
+cat(qc$system_prompt)
+```
+
+querychat doesn't need this information in any particular format -- just provide what a human would find helpful:
+
+```markdown
+# data_description.md
```
gadenbuie:
```suggestion
<!-- data_description.md -->
```

pkg-r/vignettes/data-sources.Rmd
```diff
@@ -0,0 +1,148 @@
...
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  eval = FALSE
+)
+```
+
+`querychat` supports different types of data sources, including:
```
gadenbuie:
```suggestion
`querychat` supports several different data sources, including:
```

pkg-r/vignettes/data-sources.Rmd
```diff
@@ -0,0 +1,148 @@
...
+library(palmerpenguins)
+
+querychat_app(mtcars)
+```
+
+Behind the scenes, `querychat` creates an in-memory DuckDB database and registers your data frame as a table for SQL query execution.
+
+## Database connections
+
+You can also connect `querychat` directly to any database supported by [DBI](https://dbi.r-dbi.org/). This includes popular databases like SQLite, DuckDB, PostgreSQL, MySQL, and many more.
```
gadenbuie:
```suggestion
You can also connect `querychat` directly to a table in any database supported by [DBI](https://dbi.r-dbi.org/). This includes popular databases like SQLite, DuckDB, PostgreSQL, MySQL, and many more.
```

pkg-r/vignettes/data-sources.Rmd
```diff
@@ -0,0 +1,148 @@
...
+
+Assuming you have a database set up and accessible, you can create a DBI connection and pass it to `QueryChat$new()`. Below are some examples for common databases.
```
gadenbuie: The text says `QueryChat$new()` but then the examples use `querychat_app()`. That makes sense, but should be called out to avoid confusion. Or we could switch to `querychat()` in all the places.

btw, I kind of prefer we talk about `querychat()` on the R side. Not strongly enough to say we should rewrite all of this to do that, but I think it's likely the more convenient interface for R users

pkg-r/vignettes/data-sources.Rmd
```diff
@@ -0,0 +1,148 @@
...
+con <- dbConnect(duckdb::duckdb(), dbdir = "my_database.duckdb")
+
+# Write a data frame to the database
+dbWriteTable(con, "penguins", penguins)
+
+# Or from CSV
+dbExecute(con, "
+  CREATE TABLE my_table AS
+  SELECT * FROM read_csv_auto('path/to/your/file.csv')
+")
```
gadenbuie: I'm pretty certain the duckdb R package has a friendlier function that directly does this

pkg-r/vignettes/greet.Rmd
```diff
@@ -0,0 +1,78 @@
...
+```
+
+```{r}
+library(querychat)
+library(palmerpenguins)
+```
+
+## Provide a greeting
+
+When the querychat UI first appears, you will usually want it to greet the user with some basic instructions. By default, these instructions are auto-generated every time a user arrives. In a production setting with multiple users/visitors, this is slow, wasteful, and non-deterministic. Instead, you should create a greeting file and pass it when creating your `QueryChat` object:
```
gadenbuie: > this is slow, wasteful, and non-deterministic.

This still sounds a little aggressive to me. It's _fine_ but it could be better. Or maybe it's what you want?

pkg-r/vignettes/greet.Rmd
```diff
@@ -0,0 +1,78 @@
...
+
+```{r}
+querychat_app(
```
gadenbuie: I think it'd be better to use `querychat()` here because we're talking about "in production". You could add a `qc$app()` line below, maybe in a comment, as a way to quickly test

pkg-r/vignettes/models.Rmd
```diff
@@ -0,0 +1,93 @@
...
+
+```{r}
+library(querychat)
+library(palmerpenguins)
+library(ellmer)
+```
+
+## Specify a model
+
+To use a particular model, pass a `"{provider}/{model}"` string to the `client` parameter. Under the hood, this gets passed along to `ellmer::chat()`:
```
gadenbuie: "under the hood" used twice pretty close to each other

pkg-r/vignettes/models.Rmd
```diff
@@ -0,0 +1,93 @@
...
+
+```{r}
+querychat_app(penguins, client = "anthropic/claude-sonnet-4-5")
```
gadenbuie: Again, `querychat_app()` is convenient for users to test, but I think `querychat()` is a better balance for most use cases because it can be both simple and fit in to the other more complicated examples.

In other words, `querychat()` composes with other tasks in the vignettes but `querychat_app()` is a programmatic dead-end.

pkg-r/vignettes/models.Rmd
```diff
@@ -0,0 +1,93 @@
...
+
+## Recommended models
+
+In theory, you could use any model that has tool calling support, but we currently recommend (as of November 2025):
+
+- GPT-4.1 (the default)
+- Claude 4.5 Sonnet
+- Google Gemini 3.0
+
+In our testing, we've found that those models strike a good balance between accuracy and latency. Smaller/cheaper models like GPT-4o-mini are fine for simple queries but make surprising mistakes with more complex ones; and reasoning models like o3-mini slow down responses without providing meaningfully better results.
```
gadenbuie: Interesting, I would have said all the "fast" models are pretty good, at least for most tables, i.e. I haven't had bad experiences with `gpt-4.1-mini` (even `gpt-4.1-nano` is okay) or `claude-haiku-4-5`.

I guess I'd recommend encouraging people to try out the smaller faster models first and to switch if they don't work well for the data set. (Personally, I'd be turned off from even trying the smaller models by the language "make surprising mistakes with more complex ones".)

pkg-r/vignettes/models.Rmd
```diff
@@ -0,0 +1,93 @@
...
+
+We've also seen some decent results with frontier local models, but even if you have the compute to run the largest models, they still tend to lag behind the cloud-hosted options in terms of accuracy and speed.
```
gadenbuie: Could call out `gpt-oss:20b` maybe?

pkg-r/vignettes/tools.Rmd
```diff
@@ -0,0 +1,76 @@
...
+
+Here's an example of it in action:
+
+```{r}
+querychat_app(penguins)
+```
+
+![](../reference/figures/quickstart-summary.png){alt="Screenshot of the querychat's app with a summary statistic inlined in the chat." class="shadow rounded"}
+
+## View the source
```
gadenbuie, line 64: Might need a new section for advanced use cases now that you can also use `$client()` to create a client with these tools with custom callbacks outside of a Shiny context

pkg-r/vignettes/greet.Rmd
cpsievert: 32c66dd6472458689b453ba4f5d3f7fa379868a6

pkg-r/vignettes/data-sources.Rmd
cpsievert: 32c66dd6472458689b453ba4f5d3f7fa379868a6

pkg-r/vignettes/models.Rmd
cpsievert: 32c66dd6472458689b453ba4f5d3f7fa379868a6

pkg-r/vignettes/models.Rmd
cpsievert: I didn't write this -- it is carried over from the current content. I also haven't done enough testing to have formed strong opinions about model recommendations. Do you want to take a stab at it?

pkg-r/vignettes/models.Rmd
cpsievert: I personally have experienced enough "updating of default/recommended model" fatigue that I'd generally like to avoid giving recommendations that will become outdated in a few months

pkg-r/vignettes/models.Rmd
cpsievert: 2dadb9b6254677ebbabc2aec1eedea3708ffdcd2

pkg-r/vignettes/greet.Rmd
cpsievert: 2dadb9b6254677ebbabc2aec1eedea3708ffdcd2

pkg-r/vignettes/tools.Rmd
cpsievert: Good point, but I'm not sure `tools.Rmd` is the best place for this? Anyway, since that was added after this PR was started, let's leave it for a follow up (#174)?

pkg-r/vignettes/data-sources.Rmd
cpsievert: 4fe3429598eb69fae19c8bb7c184282e1557441a

pkg-r/vignettes/models.Rmd
cpsievert: Took a minimal pass in 2a321a607ab26769c2db80780f9596048a653620

pkg-r/vignettes/context.Rmd
cpsievert: 32c66dd6472458689b453ba4f5d3f7fa379868a6

pkg-r/vignettes/build.Rmd
cpsievert: 3c9f519a5c8647d1c97df56b3f377c02d40c6e19

PR #164, opened 2025-12-10
fix(pkg-r): `querychat_app()` should only close connections if it creates the DB

gadenbuie:
Prior to this PR, if you pass in a database connection to `querychat_app()` it will close your connection for you. You only would want that if querychat created the table connection internally, i.e. if you passed in a data.frame.

## Before

```r
pkgload::load_all()

con <- DBI::dbConnect(duckdb::duckdb())
duckdb::dbWriteTable(con, "mtcars", mtcars)

querychat_app(con, "mtcars")
#> Closing data source...

DBI::dbGetQuery(con, "SELECT * FROM mtcars LIMIT 5;")
#> Error in `dbSendQuery()`:
#> ! Invalid connection
#> ℹ Context: rapi_prepare
```

## After

The database connection remains intact and the `dbGetQuery()` call completes as expected.

### DIFF
--- a/pkg-r/R/QueryChat.R
+++ b/pkg-r/R/QueryChat.R
@@ -684,6 +684,8 @@ querychat_app <- function(
   check_bool(cleanup, allow_na = TRUE)
   if (is.data.frame(data_source)) {
     cleanup <- TRUE
+  } else if (is.na(cleanup)) {
+    cleanup <- FALSE
   }
 
   qc <- QueryChat$new(


pkg-r/R/QueryChat.R
```diff
@@ -668,13 +668,22 @@ querychat_app <- function(
...
   }
 
+  if (is.na(cleanup) && interactive() && !in_shiny_session()) {
+    if (
+      is.data.frame(data_source) ||
+        inherits(data_source, "DataFrameSource")
+    ) {
+      cleanup <- TRUE
+    }
+  }
```
cpsievert: Should we do the analogous thing for Python?

gadenbuie: Yeah

gadenbuie: It turns out Python didn't have a `cleanup` parameter, so I added that to `QueryChat`. 

Python doesn't have a `querychat_app()` function, which highlights that this is a special case. I also updated the implementation slightly: we now only clean up if we're given a `data.frame` that _we_ are turning into a data source. Otherwise, calling `$cleanup()` destructively ruins a `DataSource` object, which isn't great.

pkg-py/src/querychat/_querychat_module.py
```diff
@@ -187,6 +188,13 @@ def _on_restore(x: RestoreState) -> None:
             if "querychat_has_greeted" in vals:
                 has_greeted.set(vals["querychat_has_greeted"])
 
+    if cleanup:
+        # Clean up the data source when the session ends
+
+        @session.on_ended
+        def _cleanup() -> None:
+            data_source.cleanup()
```
cpsievert: Shoot, I forgot that I intentionally hesitated on bringing the `cleanup` parameter to Python because we don't yet have the equivalent of `shiny::onStop()` -- which allows for a callback to execute on application (not session) exit. 

I think we should bring `on_stop()` to Python since it's: (1) generally useful, (2) will keep the R/Python cleanup logic in sync, and (3) I think we can have even better default behavior with it.

More specifically, for 3, I'm thinking `cleanup = TRUE` could be the default (regardless of whether you initialize in a shiny session). In that case, `Querychat()` would keep a reference to the return value of `on_stop`/`onStop()` (i.e., a callback that cancels the scheduled callback) so that it can be cancelled for `$app()`/`.app()` (but then rescheduled when that app exits).

gadenbuie: I agree about adding `on_stop()` in Shiny for Python, but in that case I'd like to revert the Python changes (or move them into another PR) so we can bring the R changes through.

More specifically, though, I think we need to go back to first principles on the `cleanup` parameter. I'm not convinced it's generally a good idea or something we should be enabling by default. 

It's very disruptive to completely shut down a connection, especially if it's a connection that we didn't create. Also, are the connections not closed automatically on garbage collection?

Anyway, I think it's reasonable for us to have a cleanup method, but I also think we should be very limited in when we call it by default, which to me is either:

1. When we create the connection and there's a clear lifespan, i.e. if we are given a data frame and create an in-memory database for `querychat_app()` (safe to shut down when exiting `querychat_app()`).
2. When the QueryChat object is created inside a Shiny app and the Shiny app ends.

If the connection is created inside an app session, we'd probably also want to shut it down, but there isn't a clear boundary of responsibilities there, so I think we're better off letting users decide

cpsievert: > but in that case I'd like to revert the Python changes (or move them into another PR) so we can bring the R changes through.

I'm OK with doing that along with a note/issue to add it later when we're happy with the R implementation.

> It's very disruptive to completely shut down a connection, especially if it's a connection that we didn't create.

Agreed, but there is also a tradeoff to consider. By restricting cleanup to `in_shiny_session()`, there's definitely going to be apps in production that don't cleanup properly. 

> Also, are the connections not closed automatically on garbage collection?

From a quick search, it doesn't seem to be a guarantee.

> I also think we should be very limited in when we call it by default

My main concern is that this will lead to complex rules around when auto-cleanup happens. And some of those rules (e.g., `interactive()`) might only make sense for one language. 

> we need to go back to first principles on the cleanup parameter

Yea, what if `cleanup = NA` resolved to `TRUE` if the number of connections grows when initializing `QueryChat()`? That way we avoid the "closing a connection we didn't open" problem and we'd also handle cleanup properly in the `not in_shiny_session()` case (well, at least for data frames -- database connections would still require a `cleanup = TRUE`)

gadenbuie: > My main concern is that this will lead to complex rules around when auto-cleanup happens. And some of those rules (e.g., `interactive()`) might only make sense for one language.

I think the rules should be simple and we don't need to rely on `interactive()`. It's in the code but I haven't been mentioning it in this conversation because it's not necessary. I think the complexity is forced on us a bit with how the on stop callback in Shiny for R can be registered globally _outside a specific app_, which likely won't be the case for Python (since it'd be a method on `session`).

I think it might be useful to reframe the problem:

1. We want to make it easy/automatic for querychat to clean up after itself when used in a Shiny app.
2. Where querychat was previously tightly coupled with use in Shiny, we now have a redesigned `QueryChat` class that is much more useful outside of a Shiny app.

The problem I want to avoid is that we take destructive actions by accident when we _aren't in a Shiny app_. I'll use R as an example:

```r
# Connect to a database (I'm hiding details, but this is usually complicated)
con <- DBI::dbConnect(...)

# Use the connection for some things
tbl1 <- dplyr::tbl(con, "one")

# Explore another table with a querychat app
querychat_app(con, "two")

# Exit the app
# Oops, your connection is dead and so are all your other tbls
```

Inside `querychat_app()`, it's easy to do the right thing: we should only shut down connections if _we_ created them to start up the app.

In the `QueryChat` class, the right thing to do depends on what method you call:

```r
qc <- QueryChat$new(con, "two")

# This shouldn't shut down the connection...
# ...unless you're inside `runApp()` or `shiny run`.
qc$app()
```

The other big signal we have is whether or not we were inside a Shiny app context when the QueryChat object is created, in which case, we _can reasonably know_ that we should shut down the connection on app end.

---

> Yea, what if `cleanup = NA` resolved to `TRUE` if the number of connections grows when initializing `QueryChat()`? That way we avoid the "closing a connection we didn't open" problem

I'm not sure how we'd track the number of connections. And even more tricky:  querychat expects to be given a connection, so other than the data frame case, it's not creating or opening a new connection anyway.

---

Anyway, there's a good chance we're saying very similar things, or arriving at similar solutions but from different starting points.

I guess, overall, my pitch is to move the mechanism for automatic cleanup down so that each method can make a better choice, or so that users can make more localized choices

gadenbuie: > > but in that case I'd like to revert the Python changes (or move them into another PR) so we can bring the R changes through.
> 
> I'm OK with doing that along with a note/issue to add it later when we're happy with the R implementation.

The R implementation in this PR is basically what I want for querychat, so I'm going to open a new issue, move this Python code to a draft PR and force push this branch back to just R changes.

gadenbuie: https://github.com/posit-dev/querychat/issues/170

cpsievert: > how the on stop callback in Shiny for R can be registered globally outside a specific app, which likely won't be the case for Python (since it'd be a method on session).

I'm confused -- Python already has `session.on_ended` -- how would `session.on_stop` be different? I was suggesting that we bring the global aspect to Python specifically because that is useful for cleanup of global side-effects (i.e., outside of a session scope).

gadenbuie: Yeah you're right; I have no idea how we'd implement an `on_stop()` in Python and the points I'm making aren't related to how it's implemented.

I'm pointing out that the problem with setting up a global `on_stop()` callback that runs when any app stops will also run when you call `qc.app()`. We're necessarily going to run into complexity as a result (that was my point in the quoted sentence).

pkg-r/R/QueryChat.R
```diff
@@ -192,7 +192,7 @@ QueryChat <- R6::R6Class(
 
       # By default, only close automatically if a Shiny session is active
       if (is.na(cleanup)) {
-        cleanup <- !is.null(shiny::getDefaultReactiveDomain())
+        cleanup <- in_shiny_session()
```
cpsievert: Sorry to keep raising somewhat tangential things, but what if this was `shiny::isRunning()`? That way, if the `QueryChat` instance is global within an app context, we'd perform the cleanup.

pkg-r/R/QueryChat.R
```diff
@@ -668,13 +668,21 @@ querychat_app <- function(
...
 ) {
   if (is_missing(table_name) && is.data.frame(data_source)) {
     table_name <- deparse1(substitute(data_source))
   }
 
+  check_bool(cleanup, allow_na = TRUE)
+  if (is.na(cleanup)) {
+    cleanup <-
+      is.data.frame(data_source) &&
+      !in_shiny_session() &&
```
cpsievert: Is it feasible for `querychat_app()` to work inside an active session? If not, maybe it'd be better to raise early?

pkg-r/NEWS.md
```diff
@@ -1,5 +1,7 @@
 # querychat (development version)
 
+* `querychat_app()` will now only automatically clean up the data source if QueryChat creates the data source internally from a data frame. (#164)
```
cpsievert, line 3: It'd be good to also have this in the docs somewhere (perhaps in the params docs for cleanup on `?querychat_app`)?

pkg-r/R/QueryChat.R
gadenbuie: that does make sense, sure!

gadenbuie: 1cbae22c3b0dab24881d49982c236f7f112a4e40

pkg-r/R/QueryChat.R
gadenbuie: d4adb514e8d55e909acf476f69fdb3dacfe8a290

cpsievert (approved):
Great, thanks!

PR #165, opened 2025-12-10
feat(pkg-r): SQL tibble sources (TblSqlSource)

### LINKED ISSUE #51: (R) Return a `dbplyr::tbl` object from `querychat_server`

gadenbuie:
Closes #51

## How does this work?

``` r
pkgload::load_all("~/work/posit-dev/querychat/pkg-r")
#> ℹ Loading querychat
library(dplyr, warn.conflicts = FALSE)
library(dbplyr, warn.conflicts = FALSE)

con <- DBI::dbConnect(duckdb::duckdb())
duckdb::dbWriteTable(con, "mtcars", mtcars)

mtcars_db <- tbl(con, "mtcars")
```

## Simple tbl source

First, we can create a new data source from the `tbl()` object.

``` r
src <- TblSqlSource$new(mtcars_db)
(res <- src$execute_query("SELECT * FROM mtcars WHERE cyl > 4"))
#> # Source:   SQL [?? x 11]
#> # Database: DuckDB 1.4.1 [root@Darwin 25.0.0:R 4.5.2/:memory:]
#>      mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#>  1  21       6  160    110  3.9   2.62  16.5     0     1     4     4
#>  2  21       6  160    110  3.9   2.88  17.0     0     1     4     4
#>  3  21.4     6  258    110  3.08  3.22  19.4     1     0     3     1
#>  4  18.7     8  360    175  3.15  3.44  17.0     0     0     3     2
#>  5  18.1     6  225    105  2.76  3.46  20.2     1     0     3     1
#>  6  14.3     8  360    245  3.21  3.57  15.8     0     0     3     4
#>  7  19.2     6  168.   123  3.92  3.44  18.3     1     0     4     4
#>  8  17.8     6  168.   123  3.92  3.44  18.9     1     0     4     4
#>  9  16.4     8  276.   180  3.07  4.07  17.4     0     0     3     3
#> 10  17.3     8  276.   180  3.07  3.73  17.6     0     0     3     3
#> # ℹ more rows
```

Which returns a `tbl()` that can be chained into further `dplyr` operations.

``` r
res |> count(cyl, gear)
#> # Source:   SQL [?? x 3]
#> # Database: DuckDB 1.4.1 [root@Darwin 25.0.0:R 4.5.2/:memory:]
#>     cyl  gear     n
#>   <dbl> <dbl> <dbl>
#> 1     6     5     1
#> 2     6     3     2
#> 3     8     3    12
#> 4     6     4     4
#> 5     8     5     2
```

## Complicated tbl source

This same process even works for more complicated tibbles, like the result of
of dplyr pipeline on SQL tibbles.

``` r
mtcars_6_8_cyl <- mtcars_db |> inner_join(mtcars_db |> dplyr::filter(cyl > 4))
#> Joining with `by = join_by(mpg, cyl, disp, hp, drat, wt, qsec, vs, am, gear,
#> carb)`
src <- TblSqlSource$new(mtcars_6_8_cyl)
```

And again, the result is a `tbl()` that can be folded into further dplyr
operations.

``` r
(res2 <- src$execute_query("SELECT * FROM mtcars_6_8_cyl WHERE gear < 6"))
#> # Source:   SQL [?? x 11]
#> # Database: DuckDB 1.4.1 [root@Darwin 25.0.0:R 4.5.2/:memory:]
#>      mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#>  1  21       6  160    110  3.9   2.62  16.5     0     1     4     4
#>  2  21       6  160    110  3.9   2.88  17.0     0     1     4     4
#>  3  21.4     6  258    110  3.08  3.22  19.4     1     0     3     1
#>  4  18.7     8  360    175  3.15  3.44  17.0     0     0     3     2
#>  5  18.1     6  225    105  2.76  3.46  20.2     1     0     3     1
#>  6  14.3     8  360    245  3.21  3.57  15.8     0     0     3     4
#>  7  19.2     6  168.   123  3.92  3.44  18.3     1     0     4     4
#>  8  17.8     6  168.   123  3.92  3.44  18.9     1     0     4     4
#>  9  16.4     8  276.   180  3.07  4.07  17.4     0     0     3     3
#> 10  17.3     8  276.   180  3.07  3.73  17.6     0     0     3     3
#> # ℹ more rows
res2 |> count(cyl, gear)
#> # Source:   SQL [?? x 3]
#> # Database: DuckDB 1.4.1 [root@Darwin 25.0.0:R 4.5.2/:memory:]
#>     cyl  gear     n
#>   <dbl> <dbl> <dbl>
#> 1     6     3     2
#> 2     8     5     2
#> 3     6     4     4
#> 4     6     5     1
#> 5     8     3    12
```

The way we make this work is by extracting the SQL for the dplyr pipeline up
until we create a data source, and then, for complicated queries at least, we
use a local CTE, letting the LLM write queries against that CTE as if it were
a fixed table.

``` r
src$complete_query("SELECT * FROM mtcars_6_8_cyl WHERE gear < 6") |> cat()
#> Error in cat(src$complete_query("SELECT * FROM mtcars_6_8_cyl WHERE gear < 6")): attempt to apply non-function
```

Amazingly, we can even apply this strategy to get the schema of the CTE. This
took a small amount of updating to `get_schema_impl()` to make it work, but
the core logic is exactly the same.

``` r
src$get_schema() |> cat()
#> Table: mtcars_6_8_cyl
#> Columns:
#> - mpg (FLOAT)
#>   Range: 10.4 to 21.4
#> - cyl (FLOAT)
#>   Range: 6 to 8
#> - disp (FLOAT)
#>   Range: 145 to 472
#> - hp (FLOAT)
#>   Range: 105 to 335
#> - drat (FLOAT)
#>   Range: 2.76 to 4.22
#> - wt (FLOAT)
#>   Range: 2.62 to 5.424
#> - qsec (FLOAT)
#>   Range: 14.5 to 20.22
#> - vs (FLOAT)
#>   Range: 0 to 1
#> - am (FLOAT)
#>   Range: 0 to 1
#> - gear (FLOAT)
#>   Range: 3 to 5
#> - carb (FLOAT)
#>   Range: 1 to 8
```

### DIFF
--- a/pkg-r/R/QueryChat.R
+++ b/pkg-r/R/QueryChat.R
@@ -142,9 +142,9 @@ QueryChat <- R6::R6Class(
     #'
     #' # With database
     #' library(DBI)
-    #' conn <- dbConnect(RSQLite::SQLite(), ":memory:")
-    #' dbWriteTable(conn, "mtcars", mtcars)
-    #' qc <- QueryChat$new(conn, "mtcars")
+    #' con <- dbConnect(RSQLite::SQLite(), ":memory:")
+    #' dbWriteTable(con, "mtcars", mtcars)
+    #' qc <- QueryChat$new(con, "mtcars")
     #' }
     initialize = function(
       data_source,
@@ -748,9 +748,9 @@ QueryChat <- R6::R6Class(
 #'
 #' # Chat with a database table (table_name required)
 #' library(DBI)
-#' conn <- dbConnect(RSQLite::SQLite(), ":memory:")
-#' dbWriteTable(conn, "mtcars", mtcars)
-#' querychat_app(conn, "mtcars")
+#' con <- dbConnect(RSQLite::SQLite(), ":memory:")
+#' dbWriteTable(con, "mtcars", mtcars)
+#' querychat_app(con, "mtcars")
 #'
 #' # Create QueryChat class object
 #' qc <- querychat(mtcars)
--- a/pkg-r/R/DBISource.R
+++ b/pkg-r/R/DBISource.R
@@ -9,11 +9,11 @@
 #' @examples
 #' \dontrun{
 #' # Connect to a database
-#' conn <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
-#' DBI::dbWriteTable(conn, "mtcars", mtcars)
+#' con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
+#' DBI::dbWriteTable(con, "mtcars", mtcars)
 #'
 #' # Create a DBI source
-#' db_source <- DBISource$new(conn, "mtcars")
+#' db_source <- DBISource$new(con, "mtcars")
 #'
 #' # Get database type
 #' db_source$get_db_type()  # Returns "SQLite"
@@ -40,9 +40,9 @@ DBISource <- R6::R6Class(
     #' @return A new DBISource object
     #' @examples
     #' \dontrun{
-    #' conn <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
-    #' DBI::dbWriteTable(conn, "iris", iris)
-    #' source <- DBISource$new(conn, "iris")
+    #' con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
+    #' DBI::dbWriteTable(con, "iris", iris)
+    #' source <- DBISource$new(con, "iris")
     #' }
     initialize = function(conn, table_name) {
       if (!inherits(conn, "DBIConnection")) {
--- a/pkg-r/R/TblSqlSource.R
+++ b/pkg-r/R/TblSqlSource.R
@@ -45,8 +45,8 @@ TblSqlSource <- R6::R6Class(
     #'   string, or will be inferred from the `tbl` argument, if possible.
     #' @return A new TblSqlSource object
     #' @examplesIf rlang::is_interactive() && rlang::is_installed("dbplyr") && rlang::is_installed("dplyr") && rlang::is_installed("RSQLite")
-    #' conn <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
-    #' DBI::dbWriteTable(conn, "mtcars", mtcars)
+    #' con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
+    #' DBI::dbWriteTable(con, "mtcars", mtcars)
     #' source <- TblSqlSource$new(dplyr::tbl(con, "mtcars"))
     initialize = function(tbl, table_name = missing_arg()) {
       check_installed("dbplyr")
--- a/pkg-r/man/DBISource.Rd
+++ b/pkg-r/man/DBISource.Rd
@@ -11,11 +11,11 @@ execution against a single table in the database.
 \examples{
 \dontrun{
 # Connect to a database
-conn <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
-DBI::dbWriteTable(conn, "mtcars", mtcars)
+con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
+DBI::dbWriteTable(con, "mtcars", mtcars)
 
 # Create a DBI source
-db_source <- DBISource$new(conn, "mtcars")
+db_source <- DBISource$new(con, "mtcars")
 
 # Get database type
 db_source$get_db_type()  # Returns "SQLite"
@@ -32,9 +32,9 @@ result <- db_source$execute_query("SELECT * FROM mtcars WHERE mpg > 25")
 ## ------------------------------------------------
 
 \dontrun{
-conn <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
-DBI::dbWriteTable(conn, "iris", iris)
-source <- DBISource$new(conn, "iris")
+con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
+DBI::dbWriteTable(con, "iris", iris)
+source <- DBISource$new(con, "iris")
 }
 }
 \section{Super class}{
@@ -78,9 +78,9 @@ A new DBISource object
 \subsection{Examples}{
 \if{html}{\out{<div class="r example copy">}}
 \preformatted{\dontrun{
-conn <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
-DBI::dbWriteTable(conn, "iris", iris)
-source <- DBISource$new(conn, "iris")
+con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
+DBI::dbWriteTable(con, "iris", iris)
+source <- DBISource$new(con, "iris")
 }
 }
 \if{html}{\out{</div>}}
--- a/pkg-r/man/QueryChat.Rd
+++ b/pkg-r/man/QueryChat.Rd
@@ -86,9 +86,9 @@ qc <- QueryChat$new(
 
 # With database
 library(DBI)
-conn <- dbConnect(RSQLite::SQLite(), ":memory:")
-dbWriteTable(conn, "mtcars", mtcars)
-qc <- QueryChat$new(conn, "mtcars")
+con <- dbConnect(RSQLite::SQLite(), ":memory:")
+dbWriteTable(con, "mtcars", mtcars)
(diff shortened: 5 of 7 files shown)


pkg-r/R/DataSource.R
```diff
@@ -378,6 +373,182 @@ DBISource <- R6::R6Class(
...
+
+      # Get the exact table name, if tbl directly references a single table
+      remote_name <- dbplyr::remote_name(private$tbl)
+
+      use_cte <- FALSE
+
+      if (!is_missing(table_name)) {
+        check_sql_table_name(table_name)
+        self$table_name <- table_name
+        use_cte <- identical(table_name, remote_name %||% remote_table)
```
cpsievert: `remote_table` isn't defined

cpsievert: Also, is the logic here inverted? That is, should it be `use_cte <- !identical(table_name, remote_name)`?

pkg-r/R/QueryChat.R
```diff
@@ -338,8 +340,15 @@ QueryChat <- R6::R6Class(
         })
 
         output$dt <- DT::renderDT({
+          df <- qc_vals$df()
+          if (inherits(df, "tbl_sql")) {
+            # Materialize the query to get a data frame, {dplyr} guaranteed by
+            # TblLazySource interface
+            df <- dplyr::collect(df)
+          }
```
cpsievert, line 436: What if we added `$tbl()` to the `DataSource` base class so that `$df()` and `$tbl()` return consistent values (i.e., `$df()` always returns a `tibble()` and `$tbl()` always returns a `tbl()`)?

pkg-r/R/DataSource.R
```diff
@@ -378,6 +373,182 @@ DBISource <- R6::R6Class(
...
+#' # Or collect the entire data frame into local memory
+#' dplyr::collect(result)
+#'
+#' # Finally, clean up when done with the database (closes the DB connection)
+#' mtcars_source$cleanup()
+#'
+#' @export
+TblLazySource <- R6::R6Class(
+  "TblLazySource",
+  inherit = DBISource,
```
cpsievert: This might be a stupid question, but assuming we have this, is there much of any reason for us also maintaining `DBISource` (i.e., maybe we deprecate it?)

pkg-r/R/DataSource.R
```diff
@@ -378,6 +373,182 @@ DBISource <- R6::R6Class(
...
+        cli::cli_abort(
+          "{.arg tbl} must be a lazy tibble connected to a database, not {.obj_type_friendly {tbl}}"
+        )
+      }
+
+      private$conn <- dbplyr::remote_con(tbl)
+      private$tbl <- tbl
+
+      # Collect various signals to infer the table name
+      obj_name <- deparse1(substitute(tbl))
```
cpsievert: I think we should be making `table_name` required and keep any `substitute()` magic further up in the call stack.

Somewhat relatedly, do you see any utility to exporting the `DataSource` implementations (i.e., `DataFrameSource`, etc)?

pkg-r/R/DataSource.R
```diff
@@ -378,6 +373,182 @@ DBISource <- R6::R6Class(
...
+      if (is.null(private$tbl_cte)) {
+        return(query)
+      }
+
+      sprintf(
+        "WITH %s AS (\n%s\n)\n%s",
+        DBI::dbQuoteIdentifier(private$conn, self$table_name),
+        private$tbl_cte,
+        query
+      )
```
cpsievert: Very clever! 💯

pkg-r/R/DataSource.R
cpsievert: Oh, now that I read the issue more closely, Oracle and SQL server don't support CTEs, so we'd mainly keep it for that reason?

pkg-r/R/QueryChat.R
```diff
@@ -338,8 +340,15 @@ QueryChat <- R6::R6Class(
         })
 
         output$dt <- DT::renderDT({
+          df <- qc_vals$df()
+          if (inherits(df, "tbl_sql")) {
```
gadenbuie, line 433: Check that `data_source` is a `TblLazySource` instead of sniffing the df result

gadenbuie: All of this isn't really _necessary_ -- I've included it because dplyr is in suggests and just want to make sure we've gone past a `check_installed("dplyr")` (which happens if you've created a `TblLazySource`).

pkg-r/R/QueryChat.R
cpsievert: After talking this through in person, I'm OK with the current approach

pkg-r/R/DataSource.R
gadenbuie: I generally agree, but in this case we're not relying _entirely_ on `substitute()`. That's actually a lower-ranked signal; we're inspecting the `tbl_sql` object directly to see if there's an associated table and if not we then fall back to using substitute. So the fact that we're relying on `tbl_sql` details means that we have to do this magic here in the class and not at a higher level.

Also I do think it's useful to export the source implementations. In most cases users won't need to access them, but it leaves the door open for specific customizations to the source objects if our input -> upgraded source logic makes any assumptions

PR #168, opened 2025-12-12
feat: Allow QueryChat usage outside of Shiny app

### LINKED ISSUE #117: Provide a way to disable analytical queries
Perhaps with something like a `tools` parameter. The benefit being that you can prevent the LLM from seeing any SQL results

### LINKED ISSUE #159: Access chat client outside of Shiny app
Hi, thanks for all the recent improvements to this package! 

I was wondering if it is possible to access the ellmer chat object created by `QueryChat$new()` after the system prompt and all querychat tooling has been added? 

My use case is that I would like to pass this as the solver in `vitals::Task$new()` to programmatically evaluate the performance of the chatbot across different models.

Thanks!

gadenbuie:
Fixes #159
Fixes #117

- New `client()` method - QueryChat objects can now create standalone chat clients with configurable tools and callbacks, enabling use outside of Shiny applications.

- New `console()` method - Launch interactive console-based chat sessions with your data source, with persistent conversation state across invocations (seems like a feature we might want to bring to the `app()` method, but there it's a little more complicated, so that can be follow-up).

- Configurable tool selection - This happened to be easy to fold into this change. A new `tools` parameter allows selective inclusion of "query" and/or "update" tools, enabling privacy-focused workflows that prevent the LLM from accessing data. 

    * Enabled tools are set when creating the class instance and the value is respected in all methods except `console()`
    * `console()` defaults to using `tools = "query"`, since the `update` tool doesn't make sense in that context
    * `client()` defaults to the instance's `tools` value but can be overridden directly.
    * All the app-related methods call the `client()` method with the default: use the instance setting for `tools`

### DIFF
--- a/pkg-r/tests/testthat/_snaps/querychat_tools.md
+++ b/pkg-r/tests/testthat/_snaps/querychat_tools.md
@@ -9,10 +9,10 @@
 ---
 
     Code
-      tool_update_dashboard(df_source, udpate_fn = NULL)
+      tool_update_dashboard(df_source, update_fn = NULL)
     Condition
       Error in `tool_update_dashboard()`:
-      ! unused argument (udpate_fn = NULL)
+      ! `update_fn` must be a function, not `NULL`.
     Code
       tool_update_dashboard(df_source, update_fn = function(query) { })
     Condition
--- a/pkg-r/tests/testthat/test-querychat_tools.R
+++ b/pkg-r/tests/testthat/test-querychat_tools.R
@@ -3,7 +3,7 @@ test_that("tool_update_dashboard() checks inputs", {
 
   df_source <- local_data_frame_source(new_test_df())
   expect_snapshot(error = TRUE, {
-    tool_update_dashboard(df_source, udpate_fn = NULL)
+    tool_update_dashboard(df_source, update_fn = NULL)
     tool_update_dashboard(df_source, update_fn = function(query) {})
     tool_update_dashboard(df_source, update_fn = function(title, extra) {})
   })


pkg-py/src/querychat/_querychat.py
```diff
@@ -46,22 +57,45 @@ def __init__(
             )
 
         self.id = id or table_name
+        self.tools = tools or ["update", "query"]
```
cpsievert: I'd expect `[]` to give me no tools, but since it's falsy, this would currently give "all tools"

pkg-py/src/querychat/_querychat.py
```diff
@@ -241,6 +275,170 @@ def generate_greeting(self, *, echo: Literal["none", "output"] = "none"):
         client.set_turns([])
         return str(client.chat(GREETING_PROMPT, echo=echo))
 
+    def _assemble_system_prompt(
```
cpsievert: Did you intend to also get rid of `def assemble_system_prompt`? 

BTW, I feel like this doesn't necessarily have to be a method: it could take just `tools`, `data_source`, and `prompt_parts` as input params.

pkg-py/src/querychat/_querychat.py
```diff
@@ -46,22 +57,45 @@ def __init__(
...
+        # Store prompt components for lazy assembly
+        if prompt_template is None:
+            prompt_template = Path(__file__).parent / "prompts" / "prompt.md"
+        self._prompt_template = (
+            prompt_template.read_text()
+            if isinstance(prompt_template, Path)
+            else prompt_template
         )
 
+        self._prompt_parts = {
```
cpsievert: Consider pulling the prompt template/parts logic out into a separate class. The init could handle this logic and it could also have a method for assembling the prompt.

pkg-py/src/querychat/_querychat.py
```diff
@@ -286,15 +484,55 @@ class QueryChat(QueryChatBase):
...
+    **Interactive console:**
+    ```python
+    from querychat import QueryChat
+    import pandas as pd
+
+    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+    qc = QueryChat(df, "my_data")
+
+    # Start an interactive console chat (query-only by default)
+    qc.console()
```
cpsievert, line 488:
```suggestion

    # Start an interactive console chat
    qc.console()
```

pkg-py/src/querychat/_querychat.py
```diff
@@ -241,6 +275,170 @@ def generate_greeting(self, *, echo: Literal["none", "output"] = "none"):
...
+                "has_tool_query": "query" in tools if tools else False,
+            },
+        )
+
+    def client(
+        self,
+        *,
+        tools: Optional[Sequence[TOOL_GROUPS] | MISSING_TYPE] = MISSING,
+        update_dashboard=None,
+        reset_dashboard=None,
```
cpsievert: Some types would be helpful here?

pkg-py/src/querychat/_querychat.py
```diff
@@ -252,7 +450,7 @@ def system_prompt(self) -> str:
             The system prompt string.
 
         """
-        return self._client.system_prompt or ""
+        return self._assemble_system_prompt()
```
cpsievert: This doesn't account for `self.tools`

pkg-py/src/querychat/_querychat.py
cpsievert: Somewhat related, `.client()` takes a different approach with `MISSING` -- let's be consistent about type/meaning here.

Also, what both of them looked something like this? 

```python
tools: TOOL_GROUPS | tuple[TOOL_GROUPS, ...] = ("update", "query")
```

And then `self.tools` gets upgraded to always be a `tuple`?

pkg-py/src/querychat/_querychat.py
cpsievert: Update: see also this comment https://github.com/posit-dev/querychat/pull/168/files#r2620999224

pkg-py/src/querychat/_querychat.py
gadenbuie: Agreed, I think I added `MISSING` later in the process and missed returning here. And yeah, we can do it as suggested

pkg-py/src/querychat/_querychat.py
gadenbuie: It does, that's why it's a method. But you're right that we could just pass in the value of `self.tools` and keep `assemble_system_prompt()` as a helper function.

Or move it into an internal `QueryChatPrompt` class to organize the prompt parts. (I'll probably head that direction.)

pkg-py/src/querychat/_querychat_module.py
```diff
@@ -91,26 +92,35 @@ def mod_server(
...
+    def reset_dashboard():
+        sql.set(None)
+        title.set(None)
 
-    # Register tools with annotations for the UI
-    chat.register_tool(update_dashboard_tool)
-    chat.register_tool(query_tool)
-    chat.register_tool(reset_dashboard_tool)
+    # Set up the chat object for this session
+    # Support both a callable that creates a client and legacy instance pattern
```
cpsievert, line 112: I don't think we need to support the legacy pattern?

pkg-py/src/querychat/_querychat.py
cpsievert: Ahh right, my bad, and sounds good.

pkg-py/src/querychat/_querychat.py
gadenbuie: cd66b0525244d905035f8d91a6fa47ffcbd59859

pkg-py/src/querychat/_querychat.py
gadenbuie: 0750ce704628e9eb8bfe7db1821ccc1fb82df223

pkg-py/src/querychat/_querychat.py
gadenbuie: Done in both Python (50760c611573adab6bcae84a425172c1cd1eb559) and R (dea193bc8924ed42ace910eff7376c032cdb3d0c)

pkg-py/src/querychat/_querychat.py
```diff
@@ -32,6 +44,7 @@ def __init__(
         id: Optional[str] = None,
         greeting: Optional[str | Path] = None,
         client: Optional[str | chatlas.Chat] = None,
+        tools: TOOL_GROUPS | tuple[TOOL_GROUPS, ...] | None | MISSING_TYPE = MISSING,
```
cpsievert: I might be missing something, but I like the approach you took in the R package of having `("update", "query")` as the default value. That's also why I suggested a tuple -- unlike lists, tuples are immutable, so they are safe to use as default value

pkg-py/src/querychat/_querychat.py
```diff
@@ -731,3 +933,40 @@ def assemble_system_prompt(
             "extra_instructions": extra_instructions_str,
         },
     )
+
+
+def _normalize_tools(
```
cpsievert: Since were in `_querychat.py`, this doesn't have to start with a `_`

pkg-py/src/querychat/_querychat.py
```diff
@@ -731,3 +933,40 @@ def assemble_system_prompt(
...
+    tools: TOOL_GROUPS | tuple[TOOL_GROUPS, ...] | None | MISSING_TYPE,
+    default: tuple[TOOL_GROUPS, ...] | None,
+) -> tuple[TOOL_GROUPS, ...] | None:
+    """
+    Normalize tools parameter to a tuple or None.
```
cpsievert: I'd be open to removing this docstring since it's pretty obvious what it's doing

pkg-py/src/querychat/_querychat.py
```diff
@@ -731,3 +933,40 @@ def assemble_system_prompt(
             "extra_instructions": extra_instructions_str,
         },
     )
+
+
+def _normalize_tools(
+    tools: TOOL_GROUPS | tuple[TOOL_GROUPS, ...] | None | MISSING_TYPE,
+    default: tuple[TOOL_GROUPS, ...] | None,
```
cpsievert, line 938:
```suggestion
    default: tuple[TOOL_GROUPS, ...],
```

pkg-py/src/querychat/_querychat.py
```diff
@@ -241,6 +261,160 @@ def generate_greeting(self, *, echo: Literal["none", "output"] = "none"):
...
+        qc.console(new=True, tools="query")
+
+        # Continue previous console session
+        qc.console()  # picks up where you left off
+        ```
+
+        """
+        # Default to query-only for console (privacy)
+        if isinstance(tools, MISSING_TYPE) and (new or self._client_console is None):
+            tools = ("query",)
```
cpsievert: Query is the less private tool?

cpsievert: Also, it's not obvious to me why `.console()` should have different default tools from `.app()`

pkg-r/R/QueryChatSystemPrompt.R
```diff
@@ -0,0 +1,106 @@
...
+
+      whisker::whisker.render(self$template, context)
+    }
+  )
+)
+
+# Utility function for loading file or string content
+read_text <- function(x) {
+  if (file.exists(x)) {
+    readLines(x, warn = FALSE) |> paste(collapse = "\n")
```
cpsievert: The old version of this was using `read_utf8()`

pkg-py/src/querychat/_querychat.py
gadenbuie: Oh right, I forgot that tuples were immutable. So yeah, happy to make that change.

pkg-py/src/querychat/_querychat.py
gadenbuie: oops, no that comment isn't right, it's actually the opposite. I'll fix that.

`.console()` needs different tools from `.app()` because the update dashboard tool doesn't make sense in the console where there's no dashboard to update. Really, only the query tool makes sense and it's safe for that to be the default because this for interactive use, not something you'd put into production.

You _could_ go through hoops to set up `update_dashboard` in a way that it works or makes sense for something you're working on, but `tools = "query"` is the right default choice for `.console()`.

pkg-r/R/QueryChatSystemPrompt.R
gadenbuie: good catch thanks

pkg-py/src/querychat/_querychat.py
gadenbuie: `default` needs to be able to take `None` because `self.tools` can be `None` and in cases we use `normalize_tools()` to fall back to the global default if `tools = MISSING`.

pkg-py/src/querychat/_querychat.py
gadenbuie: To be a little more concrete about the hoops: `.console()` accepts kwargs that are passed to `.client()` which you could use to setup `update_dashboard` and `reset_dashboard` callbacks.

pkg-py/src/querychat/_utils.py
```diff
@@ -11,6 +11,15 @@
...
 
+class MISSING_TYPE:  # noqa: N801
+    """
+    A singleton representing a missing value.
+    """
+
+
+MISSING = MISSING_TYPE()
+
+
```
cpsievert, line 22:
```suggestion
```

gadenbuie: 🤔 were there other suggestions? we're still using `MISSING` on the `.client()` method because we need to differentiate between `tools=None` (attach no tools) and `tools=MISSING` (use the class instance default)

pkg-py/src/querychat/_querychat.py
cpsievert: Sounds good. 

BTW, this PR does start to make me think that it'd useful to have non-Shiny/reactive versions of `.df()`, etc., that you could read directly from the instance. And in that case, tools would also now update the values that those methods return. Seems this would be somewhat useful R and also might end up being really nice for supporting streamlit, etc on the Python side. I was sort of planning on dedicating some time to streamlit this week, so maybe I'll explore this idea once this PR is done?

It does also kinda push me a direction of thinking a `.chat()` and `.stream()` method might be useful?

pkg-py/src/querychat/_querychat.py
cpsievert: Ah yea, ok.

pkg-py/src/querychat/_utils.py
cpsievert: Ohh, shoot, ok. I was more-so thinking you could removing the `MISSING_TYPE` for `tools` on `__init__`

cpsievert: Looks like it's been resolved, so nvm

pkg-py/src/querychat/_querychat.py
gadenbuie: I'm not sure about the non-reactive `.df()`, but as to `.chat()` and `.stream()` -- I created `.client()` to help us go down that path. My intuition is that in Streamlit, notebook, etc. contexts you might be able to get pretty far with superclasses that can rely on `.client()`, or maybe we'll need the `.chat()` and `.stream()` methods.

I don't think we should build any of these right this minute, but the goal of `.client()` is to make it easier to test and find out.

PR #172, opened 2025-12-16
fix: Expose `id` in ui/server methods to support Shiny for R modules and multiple instances

### LINKED ISSUE #169: Unable to run QueryChat inside shinyModule
QueryChat doesn't appear to work properly when using it inside a module. The chat spinner just spins indefinitely. 

In the examples I've found so far, the qc object is initialized at the global level. However, it works as expected when initialized in the server w/o shiny modules. My motivation for doing it this way is to be able to pass data.tables/data.frames that are created within an app. I also noticed that when using renderUI, the greeting message does not appear and am wondering if it's related to this issue. Ultimately, I think it's a name spacing issue with `mod_ui/server` but I can't track it down.

Really appreciate any help!

Reprex below
```
library(shiny)
library(bslib)
library(querychat)
library(ellmer)

options(duckdb.debug = T)

# Module UI
test_module_ui <- function(id) {
  ns <- NS(id)
  page_fluid(
    card(
      DT::DTOutput(ns("dt"))
    ),
    card(
      uiOutput(ns("querychat_sidebar"))
    )
  )
}

# Module Server
test_module_server <- function(id) {
  moduleServer(id, function(input, output, session) {
    ns <- session$ns

    message("Module ID: ", id)
    message("Session namespace: ", session$ns(NULL))

    # Create QueryChat instance
    qc <- QueryChat$new(
      mtcars,
      id = id,
      client = ellmer::chat_aws_bedrock(model = "us.anthropic.claude-sonnet-4-5-20250929-v1:0"),
      greeting = "Ask me about mtcars!"
    )

    message("QueryChat created with ID: ", qc$id)

    # Initialize the QueryChat server
    qc_vals <- qc$server()

    # Render the QueryChat UI
    output$querychat_sidebar <- renderUI({
      qc$ui()
    })

    # Render the data table
    output$dt <- DT::renderDT({
      DT::datatable(qc_vals$df())
    })
  })
}

# Main App UI
ui <- fluidPage(
  titlePanel("QueryChat Module Test"),
  test_module_ui("test_module")
)

# Main App Server
  test_module_server("test_module")
}

shinyApp(ui, server)
```

Chat hangs
Image

cpsievert:
Closes #169

This PR introduces proper support for using QueryChat within Shiny (for R) modules by allowing explicit `id` parameter overrides in `$ui()`, `$sidebar()`, and `$server()` methods.

### Key Changes

**1. Added `id` parameter to `$ui()`, `$sidebar()`, and `$server()` methods:**
   - Users can now explicitly pass an `id` when calling these methods
   - When not provided, falls back to the ID specified at initialization
   - In Shiny modules, users pass namespaced IDs using standard Shiny module patterns: `id = ns("your_id")`

**2. Follows idiomatic Shiny module patterns:**
   - Uses the standard Shiny approach where module IDs are wrapped in `ns()` in the UI function
   - The corresponding server function receives the unwrapped ID
   - This is the established pattern that Shiny developers are familiar with

**3. Enables multiple instances from the same QueryChat object:**
   - By allowing explicit `id` override, users can create multiple chat interfaces from a single QueryChat instance
   - Useful for both module and non-module scenarios where multiple chats are needed

### Implementation Details

- Modified `$ui()` and `$sidebar()` to accept an optional `id` parameter instead of `ns` function
- Modified `$server()` to accept an optional `id` parameter
- When `id` is not provided, falls back to `self$id` with automatic namespace detection (via `namespaced_id()`) for UI methods
- Updated documentation to clarify the new module usage pattern
- Removed the error check that prevented setting `id` directly

## Usage

**In a Shiny module UI function:**

```r
module_ui <- function(id, qc) {
  ns <- NS(id)
  card(qc$ui(id = ns("qc-ui")))  # Standard Shiny module pattern
}
```

**In a Shiny module server function:**

```r
module_server <- function(id, qc) {
  moduleServer(id, function(input, output, session) {
    qc$server(id = "qc-ui")  # Matches the unwrapped ID from UI
  })
}
```

**In a regular Shiny app (no module):**

```r
ui <- page_fluid(
  qc$ui()  # Uses default ID from initialization
)

  qc$server()  # Uses default ID from initialization
}
```

### DIFF
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New features
 
-* `QueryChat.sidebar()`, `QueryChat.ui()`, and `QueryChat.server()` now support an optional `id` parameter to enable use within Shiny modules or to create multiple chat instances from a single `QueryChat` object. (#172)
+* `QueryChat.sidebar()`, `QueryChat.ui()`, and `QueryChat.server()` now support an optional `id` parameter to create multiple chat instances from a single `QueryChat` object. (#172)
 
 * `QueryChat.client()` can now create standalone querychat-enabled chat clients with configurable tools and callbacks, enabling use outside of Shiny applications. (#168)
 


pkg-r/R/QueryChat.R
gadenbuie, line 181: While we're here, I think we should prepend `querychat_` to the default `id`


```suggestion
      self$id <- id %||% sprintf("querychat_%s", table_name)
```

gadenbuie:
@cpsievert I updated the PR description, made the `"querychat_" + table_name` for default ID, and added an example app in the R package as `03-module-app`. I also made similar changes on the Python side (except for adding an example app).

pkg-py/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New features
 
+* `QueryChat.sidebar()`, `QueryChat.ui()`, and `QueryChat.server()` now support an optional `id` parameter to enable use within Shiny modules or to create multiple chat instances from a single `QueryChat` object. (#172)
```
cpsievert: This actually isn't a problem for PyShiny modules since UI will implicitly namespace there.


```suggestion
* `QueryChat.sidebar()`, `QueryChat.ui()`, and `QueryChat.server()` now support an optional `id` parameter to create multiple chat instances from a single `QueryChat` object. (#172)
```

cpsievert:
Thanks! LGTM other than the one suggestion.

PR #173, opened 2025-12-17
fix(pkg-r): Directly expose `id` rather than ask for `ns()` function

gadenbuie:
This stacks on #172 and is an alternate approach that exposes `id`.

The `ns` argument in #172 is convenient, but it's also not a standard a pattern that we use in Shiny modules. In general, it's normal to expect that you'll need to wrap any `id` for any component in `ns()`, e.g. `id = ns("component")`, in a Shiny module, including when you're nesting Shiny modules.

Passing in `ns` is an interesting idea, but it's not the standard established pattern and I'm not sure it's entirely worth it because it doesn't entire enable another class of use cases.

Allowing users to pass in `id`, on the other hand, **is** idiomatic and expected and gives users extra flexibility to call `$ui()` and `$server()` with alternate IDs, even in a non-module use case. That's entirely valid because we've decoupled the `$server()` return values from the parent QueryChat class.

## Example

Here's the example in #172 updated to use the new pattern.

```r
library(shiny)
library(bslib)
library(querychat)

qc <- QueryChat$new(mtcars, greeting = "Ask me about mtcars!")

module_ui <- function(id, qc) {
  ns <- NS(id)
  card(qc$ui(id = ns("qc-ui"))) #<< this is normal Shiny module stuff
}

module_server <- function(id, qc) {
  moduleServer(id, function(input, output, session) {
    qc$server(id = "qc-ui") #<< also expected in a Shiny module
  })
}

ui <- page_fluid(
  titlePanel("QueryChat Module Test"),
  module_ui("module1", qc)
)

  module_server("module1", qc)
}

shinyApp(ui, server)
```

So while the new R6 class saves you from needing to pass IDs around when you're using `qc$ui()` and `qc$server()` in a normal Shiny app (and you only have one), if you're in Shiny modules or if you have more than one chat from the same `qc` source object, you'll be able to connect the `id`s by passing them explicitly.

### DIFF
--- a/pkg-r/R/QueryChat.R
+++ b/pkg-r/R/QueryChat.R
@@ -474,8 +474,10 @@ QueryChat <- R6::R6Class(
     #' @param width Width of the sidebar in pixels. Default is 400.
     #' @param height Height of the sidebar. Default is "100%".
     #' @param fillable Whether the sidebar should be fillable. Default is `TRUE`.
-    #' @param ns A Shiny namespacing (i.e., [shiny::NS()]) function.
-    #' Only needed when calling this method within a module UI function.
+    #' @param id Optional ID for the QueryChat instance. If not provided,
+    #'   will use the ID provided at initialization. If using `$sidebar()` in
+    #'   a Shiny module, you'll need to provide `id = ns("your_id")` where `ns`
+    #'   is the namespacing function from [shiny::NS()].
     #'
     #' @return A [bslib::sidebar()] UI component.
     #'
@@ -493,7 +495,7 @@ QueryChat <- R6::R6Class(
       width = 400,
       height = "100%",
       fillable = TRUE,
-      ns = NULL
+      id = NULL
     ) {
       bslib::sidebar(
         width = width,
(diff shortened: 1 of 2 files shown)


cpsievert:
> Allowing users to pass in id, on the other hand, is idiomatic and expected 

I feel a bit torn on how true this is. It is idiomatic if you know that `$ui()`/`$server()` functions are module functions, but how obvious is that? And how familiar of a pattern is it to call a module from within another module? 

> The ns argument in #172 is convenient, but it's also not a standard a pattern that we use in Shiny modules

I get where you're coming from, but it also sort of comes from a place that you know the underlying implementation of `$ui()`/`$server()` is a module. Also, the class based approach is already non-standard, so I'm not sure following standards is obviously more helpful.

Ultimately, I do feel a bit conflicted on this, but the aspect that pushed me towards `ns` was the fact that it is confusing for the instance to have an `id` that you can also override into irrelevance with `$ui()` and `$server()`. With `ns`, there is no confusion over when/which `id`s to set, and it's more obvious that it's only there because of the modules use case.

PR #175, opened 2025-12-19
feat(pkg-py): Replace pandas with narwhals

cpsievert:
This PR removes `pandas` as a required dependency and replaces it with `narwhals`, a lightweight DataFrame abstraction layer that supports both pandas and polars backends. Users can now choose their preferred DataFrame library.

## Motivation

- **Reduced dependencies**: pandas is a heavy dependency; users who prefer polars shouldn't need to install pandas
- **Flexibility**: Some users prefer polars for its performance characteristics
- **Future-proofing**: narwhals provides a stable API that works across DataFrame libraries

### Dependencies
- Removed `pandas` from required dependencies
- Added `pandas` and `polars` as optional dependencies
- Install with `pip install querychat[pandas]` or `pip install querychat[polars]`

### API Changes (Breaking)
- `execute_query()`, `get_data()`, and `df()` now return narwhals DataFrames instead of pandas DataFrames
- Call `.to_native()` on returned DataFrames to get the underlying pandas/polars DataFrame

### Internal Changes
- New `_df_compat.py` module handles backend selection (prefers polars when available)
- `df_to_html()` generates HTML directly without pandas dependency
- `DataFrameSource` accepts pandas, polars, or narwhals DataFrames

### Tests
- Added `test_df_compat.py` for the compatibility layer
- Added `test_dataframe_source.py` with comprehensive DataFrameSource tests
- Polars-specific tests are skipped when polars isn't installed

## Migration Guide

```python
# Before
df = querychat.df()  # pandas DataFrame
df.head()

# After
df = querychat.df()  # narwhals DataFrame
df.head()            # still works - narwhals has similar API

# If you need a native pandas/polars DataFrame:
native_df = querychat.df().to_native()
```

### DIFF
--- a/pkg-py/src/querychat/_querychat.py
+++ b/pkg-py/src/querychat/_querychat.py
@@ -143,7 +143,7 @@ class QueryChatBase:
                 self.id,
                 data_source=self._data_source,
                 greeting=self.greeting,
-                client=self.client(),
+                client=self.client,
                 enable_bookmarking=enable_bookmarking,
             )
 
@@ -809,7 +809,7 @@ class QueryChatExpress(QueryChatBase):
             self.id,
             data_source=self._data_source,
             greeting=self.greeting,
-            client=self.client(),
+            client=self.client,
             enable_bookmarking=enable,
         )
 


pkg-py/src/querychat/_datasource.py
```diff
@@ -99,27 +99,26 @@ def cleanup(self) -> None:
 
 
 class DataFrameSource(DataSource):
-    """A DataSource implementation that wraps a pandas DataFrame using DuckDB."""
+    """A DataSource implementation that wraps a DataFrame using DuckDB."""
 
-    _df: nw.DataFrame | nw.LazyFrame
```
cpsievert, line 138: I'm pretty sure it's going to make sense for us to have a separate `LazyFrameSource`, which I'll do in a follow up PR (before the next release). The benefit being that we can be more lazy about computation, and possibly have `.df()` also return a `LazyFrame` in that scenario

pkg-py/src/querychat/_datasource.py
```diff
@@ -151,16 +150,8 @@ def get_schema(self, *, categorical_threshold: int) -> str:
         """
         schema = [f"Table: {self.table_name}", "Columns:"]
 
-        # Ensure we're working with a DataFrame, not a LazyFrame
-        ndf = (
-            self._df.head(10).collect()
```
cpsievert, line 208: Note that downstream calculation of ranges and unique values wasn't working properly because they were based on the first 10 rows -- I'll address this when doing the new `LazyFrameSource` implementation

pkg-py/src/querychat/_utils.py
```diff
@@ -223,11 +234,30 @@ def df_to_html(df: IntoFrame, maxrows: int = 5) -> str:
...
+    # Body
+    html_parts.append("  <tbody>")
+    for row in rows:
+        html_parts.append("    <tr>")
+        html_parts.extend(f"      <td>{_escape_html(str(val))}</td>" for val in row)
+        html_parts.append("    </tr>")
+    html_parts.append("  </tbody>")
+
+    html_parts.append("</table>")
+    table_html = "\n".join(html_parts)
```
cpsievert: I feel a bit torn on whether it's worth a `great_tables` dependency here? It is pretty low dependency, so maybe it's worth it?

pkg-py/src/querychat/_datasource.py
gadenbuie: makes sense!

pkg-py/src/querychat/_utils.py
gadenbuie: I think it makes sense to use `great_tables` if it drops in nicely

gadenbuie (approved):
Nice!

PR #195, opened 2026-01-15
feat: multi-table support and data dictionary

cpsievert:
## For users

querychat has been single-table: one data frame, one schema, one conversation. If you had orders and customers and products, your options were to pre-join everything into one wide frame (losing the relational structure the LLM could reason about) or to spin up separate querychat sessions (losing the ability to ask cross-table questions). This PR removes that limitation.

You can now register multiple tables and the LLM will reason across all of them — joins, cross-table filters, aggregations:

```python
qc = QueryChat(orders_df, "orders")
qc.add_table(customers_df, "customers")
```

If your data lives in a SQLAlchemy engine (Python) or DBI connection (R), `add_tables()` bulk-registers all tables — or a named subset — in a single call, and builds the system prompt exactly once rather than once per table:

```python
qc = QueryChat()
qc.add_tables(engine)                          # all tables
qc.add_tables(engine, ["orders", "customers"]) # specific subset
```

The LLM's SQL queries can reference any registered table. To consume per-table filter state in a Shiny app, call `qc.server()` and use `table()` on the returned values — each table gets its own reactive sql/df/title:

```python
# In your Shiny server function:
qc_vals = qc.server()
qc_vals.table("orders").df()
qc_vals.table("customers").sql()
```

As a side note: `qc_vals.current_table()` (R: `qc_vals\$current_table()`) is also available — a reactive that returns the name of the most recently LLM-queried table, or `None`/`NULL` before any query. Useful when you want an output to follow whatever table the LLM just acted on without hard-coding a name.

Under the hood, a new `QueryExecutor` layer dispatches queries to the right backend. DataFrame sources (pandas, polars, pyarrow) get a shared DuckDB connection; Polars LazyFrames get a shared `SQLContext`; SQLAlchemy/Ibis tables use the existing shared backend directly. All tables in a session must be the same source type.

But multi-table introduces a scaling problem. The old design embedded full column statistics in the system prompt at startup — every conversation paid the cost of computing stats for every column of every table, whether the LLM needed them or not. With multiple tables, that cost multiplies. So schema fetching is now **on-demand**: the system prompt contains only a lightweight table overview, and the LLM calls the new `querychat_get_schema` tool to get column details for specific tables when it actually needs them.

And when you have several tables, column names get ambiguous. What does `status` mean in the orders table vs. the customers table? A new **`DataDict`** type — integrating with the https://data-dict.tidyverse.org/ spec — lets you annotate tables and columns with plain-English descriptions, loaded from a YAML file:

```python
QueryChat(data_dict="data_dict.yaml")
```

The schema tool uses these annotations directly, skipping the live stats query for any column that already has metadata. A fully-documented table incurs zero stats overhead. `DataDict` also supports `relationships` (JOIN hints for multi-table) and a `glossary` (domain term definitions), both surfaced in the system prompt so the LLM understands your domain language.

These three pieces — multi-table, on-demand schema, and data dictionary — aren't independent features. They're designed as layers: multi-table creates the need for smarter schema handling, on-demand fetching solves the cost problem, and `DataDict` solves the precision problem.

### Breaking changes

| Before | After |
|---|---|
| `qc.data_source` | `qc.table("name").data_source` |
| `qc.data_source = new_df` | `qc.add_table(new_df, "name", replace=True)` |
| `qc.server(data_source=df)` (Shiny) | `qc.add_table(df, "name")` before `server()` |
| `qc.df()` / `qc.sql()` / `qc.title()` (multi-table) | `qc_vals.table("name").df()` etc. via server return |

Bookmark keys changed: Python uses flat per-table keys (`querychat_sql_{name}`, `querychat_title_{name}`); R uses a nested `querychat_tables` dict. Old single-table flat keys (`querychat_sql`) are no longer written.

`data_description` is deprecated in favor of `data_dict`.

---

## For reviewers

The code mirrors the same layered story. Each layer depends only on the ones below it, so reviewing bottom-up avoids forward references:

**Layer 1 — The data contract** (`_datasource.py`). `DataSource` gained two methods: `get_column_metas()` (cheap — column names and types via `LIMIT 0`) and `populate_column_stats()` (expensive — min/max/categories). This split is the foundation everything else builds on: it's what makes on-demand fetching possible and what lets `DataDict` selectively skip stats queries. `ColumnMeta` also gained a `description` field for `DataDict` annotations.

**Layer 2 — Multi-table dispatch** (`_query_executor.py`, new). A `QueryExecutor` sits between the tools and the data sources. It exposes the same `get_column_metas(table)` / `populate_column_stats(table)` / `execute_query()` surface but routes to the right backend. Three implementations: `DataSourceExecutor` for single-table or shared-backend multi-table (SQLAlchemy, Ibis); `DuckDBExecutor` for multi-table DataFrames (registers all tables in one in-memory DuckDB); and `PolarsSQLExecutor` for multi-table Polars LazyFrames (shared `SQLContext`). `check_source_compatibility()` enforces that all tables in a session share the same source type and backend.

**Layer 3 — Column metadata** (`_data_dict.py`, new). A Pydantic model hierarchy: `DataDict` → `TableSpec` → `ColumnSpec`. The key method, `DataDict.get_table_schema()`, merges static annotations onto live `ColumnMeta` objects and only sends undocumented columns to `populate_column_stats()`. This is where the "skip stats for documented columns" optimization lives.

**Layer 4 — The schema tool** (`tools.py`, `prompts/tool-get-schema.md`). `querychat_get_schema` dispatches to either `DataDict.get_table_schema()` or `executor.get_schema()` depending on whether a dict is present. It returns a `GetSchemaResult` subclass that renders a compact status line in Shiny chat rather than dumping the raw schema.

**Layer 5 — Prompt rewiring** (`_system_prompt.py`, `prompts/prompt.md`). The old `{{schema}}` Mustache variable is gone. Replaced by `{{tables_overview}}` (lightweight table list), `{{relationships}}`, `{{glossary}}`, and a `{{multi_table}}` flag that drives conditional instructions requiring `table=` parameters in tool calls.

**Layer 6 — Orchestration** (`_querychat_base.py`). The base class now holds `dict[str, DataSource]`. `add_table()` uses a staged-rebuild pattern: it builds the new system prompt and executor against a copy of the sources dict, committing only if both succeed, to avoid partial state on failure. `add_tables()` (Python: SQLAlchemy engine; R: DBI connection) bulk-stages all tables and calls `build_system_prompt()` exactly once, avoiding the N-1 spurious rebuilds that calling `add_table()` in a loop would cause. `table(name)` returns a config-only `TableAccessor` — `data_source` and `table_name` work, but `df()`/`sql()`/`title()` raise with guidance to use the server return value instead.

**Layer 7 — Shiny reactive state** (`_shiny_module.py`). `ServerValues` has per-table state in `tables: dict[str, TableState]`, each with its own reactive `sql`, `title`, and `df`. It also gained a `table(name)` method that returns a `TableAccessor` backed by the per-session `TableState` — this is the correct way to consume per-table reactive results. Also added: `current_table()`, which tracks the most recently LLM-touched table name across `update_dashboard`, `reset_dashboard`, `chat_update`, and bookmark restore. Single-table apps still get top-level `.df`, `.sql`, `.title` on `ServerValues` for backward compatibility; multi-table replaces them with `_MultiTableBlockedReactive` sentinels that raise with migration guidance. Bookmark keys are per-table (Python: flat `querychat_sql_{name}`; R: nested `querychat_tables`).

**Layer 8 — Public accessors and framework adapters** (`_table_accessor.py` new, `_dash.py`, `_gradio.py`, `_streamlit.py`). `TableAccessor` is a thin proxy that can be constructed in two modes: config-only (from `qc.table()`, raises on reactive methods) or state-backed (from `qc_vals.table()`, fully reactive). `StateDictTableAccessor` is the Dash/Gradio variant that always raises on reactive methods with instructions to use the state-dict API instead. Framework adapters gained `table=` parameters on `df()`, `sql()`, `title()`.

## Test plan

- [ ] `make py-check` passes (format, types, tests)
- [ ] Multi-table Shiny app: LLM correctly joins/filters across tables
- [ ] Single-table apps still work with no API changes
- [ ] DataDict: YAML loads, enriched descriptions reach the LLM, documented columns skip stats queries
- [ ] Bookmark round-trip: new per-table keys save/restore correctly; old flat keys still load


pkg-py/CHANGELOG.md
```diff
@@ -9,10 +9,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### New features
 
+* `QueryChat()` now supports **multiple related tables**. Register additional tables with `add_table()` and the LLM can reason across all of them — joins, cross-table filters, aggregations. Per-table reactive state (`df()`, `sql()`, `title()`) is accessible via `qc.table("name")`. (#195)
+
+  ```python
+  qc = QueryChat(orders_df, "orders")
+  qc.add_table(customers_df, "customers")
+
+  qc.table("orders").df()
+  qc.table("customers").sql()
```
cpsievert: This should be `qc_vals`, not `qc` (see the PR description)

pkg-py/CHANGELOG.md
```diff
@@ -9,10 +9,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  ```
+
+* A new **`DataDict`** type lets you annotate columns with plain-English descriptions loaded from a YAML file. The LLM receives these descriptions when it fetches the schema, helping it interpret ambiguous or domain-specific column names without any extra prompting. (#195)
+
+  ```python
+  qc = QueryChat(df, "sales", data_dict=DataDict.from_yaml("schema.yaml"))
```
cpsievert:
```suggestion
  QueryChat(data_dict="data_dict.yaml")
```

pkg-py/CHANGELOG.md
```diff
@@ -9,10 +9,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+  ```python
+  qc = QueryChat(orders_df, "orders")
+  qc.add_table(customers_df, "customers")
+
+  qc.table("orders").df()
+  qc.table("customers").sql()
+  ```
+
+* A new **`DataDict`** type lets you annotate columns with plain-English descriptions loaded from a YAML file. The LLM receives these descriptions when it fetches the schema, helping it interpret ambiguous or domain-specific column names without any extra prompting. (#195)
```
cpsievert: Link to the actual project (Again, see the PR description)

pkg-py/CHANGELOG.md
```diff
@@ -9,10 +9,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+  ```python
+  qc = QueryChat(df, "sales", data_dict=DataDict.from_yaml("schema.yaml"))
+  ```
+
+* The system prompt is now lighter: full schema is no longer embedded upfront. Instead the LLM fetches per-table schema on demand via the new `querychat_get_schema` tool — and only when it needs to. When a `DataDict` is provided, the tool skips columns that already have descriptions, so the LLM only pays for what isn't already documented. (#195)
```
cpsievert: This is more of an improvement than it is a new feature.

pkg-py/CHANGELOG.md
cpsievert: Also mentioned that this is the preferred way to provide additional context for the data, especially when multiple tables are relevant.

pkg-py/CHANGELOG.md
```diff
@@ -83,6 +106,10 @@ Each framework's `QueryChat` provides `.app()` for quick standalone apps and `.u
 
 ### New features
 
+* Added multi-table support. Use `qc.add_table(df, "table_name")` to register additional tables after construction, then write cross-table SQL (JOINs, subqueries) in queries. Access per-table state via `qc.table("name")`. All tables must share the same backend type (e.g., all pandas DataFrames or all SQLAlchemy tables on the same engine).
+
+* Added `PolarsLazySource` to support Polars LazyFrames as data sources. Data stays lazy until the render boundary, enabling efficient handling of large datasets. Pass a `polars.LazyFrame` directly to `QueryChat()` and queries will be executed lazily via Polars' SQLContext.
```
cpsievert: I think this section needs to be removed?

pkg-py/CHANGELOG.md
```diff
@@ -9,10 +9,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 * File attachments are now enabled by default in the Shiny chat UI. Users can attach images, PDFs, and text files to their messages and the LLM will receive them. Disable with `allow_attachments=False` in `mod_ui()` or `QueryChat.ui()`. (#253)
 
+### Breaking Changes
+
+* The `data_source` property has been removed. Use `qc.table("name").data_source` to read a table's data source, and `qc.add_table(df, "name", replace=True)` to replace it. The `data_source` parameter to `server()` (Shiny) has also been removed; call `add_table()` before `server()` instead. (#195)
+
 ### Improvements
 
+* `data_description` is superseded by `data_dict` for new code; existing usage continues to work without any warnings. (#195)
```
cpsievert: Could probably remove this after improving the other point about the data dictionary.

gadenbuie (commented):
First off — this is a lot of well-structured work, and the layered design (data contract → executor → data dict → tools → prompt → orchestration → reactive state → adapters) made it genuinely reviewable despite its size. The notes below are about composition and where we can trim incidental complexity; none are merge-blockers.

**At a glance:**

1. **Coalesce more source types into a shared DuckDB backend** — pins + dataframes
   already both materialize into DuckDB; the class-identity compatibility check is
   the only thing stopping them (and other blends) from sharing one connection.
   Forward-looking; biggest design idea.
2. **Better affordances for registering many tables from a database** — connecting
   an existing multi-table DB is fully manual today; suggest auto-discovery on
   construction and a DB-backed multi-table example.
3. **Simplification opportunities within the PR as-is** — three concrete,
   mostly-mechanical refactors (a mislayered mixin, a duplicated back-compat rule,
   a partly-duplicated executor split) that reduce complexity without changing
   behavior. These are *recommendations* — stronger than "nice to have," not
   merge-blocking.

Sections 1 and 2 are forward-looking direction. Section 3 is the actionable near-term feedback.

---

## 1. Coalesce more source types into a shared DuckDB backend

### Motivation

The PR enforces that all tables in a session share the same source *type* and backend (`check_source_compatibility` in `_query_executor.py`, gating on `type(new_source) is not type(first_source)`). That rejects natural blends — e.g. a pin combined with a local data frame, or a database table combined with a CSV.

Worth emphasizing: blending heterogeneous data sources like this is likely to be a *common* ask, not an edge case — and being able to join a pin, a local data frame, and a database table in one conversation is a genuine strength of the R and Python ecosystems that is hard to replicate in other languages. This is worth leaning into as a differentiator, not just accommodating.

But several of these sources already materialize into DuckDB independently:

- `DataFrameSource` — own private `duckdb.connect()`, registers the frame, locks down.
- `PinSource` — own private `duckdb.connect()`, loads the pin into a DuckDB table, locks down.

Both expose `get_data()` returning a native dataframe, and both report `get_db_type() == "DuckDB"`. The multi-table `DuckDBExecutor` is *already* a generic "register dataframes into one shared in-memory DuckDB" engine:

```python
# _query_executor.py — DuckDBExecutor.__init__
for name, source in sources.items():
    self._conn.register(name, source.get_data())
```

It consumes nothing `DataFrameSource`-specific. A `PinSource` satisfies the same contract today. So the thing blocking pins + dataframes in one session is the class-identity check, not the engine.

### The unifying idea: invert the contract

Instead of executors knowing about specific source classes, have each source
declare **how it contributes itself to a shared DuckDB connection**, and let one
generalized executor drive them all.

Proposed contract method on `DataSource` (name TBD):

```python
def register_into(self, conn: duckdb.DuckDBPyConnection, *, profile: LockdownProfile) -> None:
    ...
```

Per-source strategies — this table *is* the design:

| Source | `register_into` strategy | Cost |
|---|---|---|
| `DataFrameSource` | `conn.register(name, self._df.to_native())` | zero-copy view |
| `PinSource` | `conn.register(name, self.get_data().to_native())` | already-materialized frame |
| (new) file source | `CREATE VIEW name AS SELECT * FROM read_parquet(path)` | lazy, on-disk |
| `SQLAlchemySource` (PG/MySQL/SQLite) | `ATTACH '<connstr>' AS x (TYPE postgres, READ_ONLY); CREATE VIEW name AS SELECT * FROM x.schema.tbl` | live |
| `SQLAlchemySource` (same, snapshot) | `CREATE TABLE name AS SELECT * FROM x.schema.tbl` | startup copy |
| `IbisSource` / Snowflake / Databricks | pull to Arrow, then `conn.register` | snapshot via Python |

A source that can't (or shouldn't) register simply doesn't implement the contract and falls back to today's `DataSourceExecutor`. `check_source_compatibility` then becomes a **partition** ("can these share one DuckDB?") rather than a hard reject on class identity.

### Known limitation: DuckDB attach scanners are DB-specific

DuckDB's database attach scanners exist only for **Postgres, MySQL, and SQLite**. But `SQLAlchemySource` today advertises Postgres, MySQL, SQLite, **Snowflake, and Databricks** (`_datasource.py:524-525`), plus there's a separate `IbisSource`.

- PG / MySQL / SQLite → can `ATTACH` live into DuckDB. Full win.
- Snowflake / Databricks / arbitrary Ibis → no DuckDB scanner. The only way to
  coalesce is to pull through the Python client into Arrow and register a
  **snapshot**, which forfeits live querying and can be large. For these, "stay
  in the native live executor" is the honest answer.

Also note: building DuckDB's `ATTACH` string from a SQLAlchemy `engine.url` is real per-dialect glue — DuckDB wants libpq-style `dbname=... host=... user=...`, not SQLAlchemy's `postgresql+psycopg2://`.

### Lockdown profiles

Today `duckdb_lock_down` (`_datasource.py:172-182`) is all-or-nothing and sets `enable_external_access = false` — exactly what blocks a live remote attach. Supporting live remote queries means turning the lockdown into a spectrum:

1. **`sealed`** (today's behavior) — no external access, local FS disabled, config
   locked. Dataframes / pins / snapshots only.
2. **`attached_readonly`** — `enable_external_access = true` (required for attach to
   function), but: every remote DB attached `READ_ONLY`; `disabled_filesystems =
   'LocalFileSystem'` still set so LLM SQL can't read local files; extension
   autoload off; `lock_configuration = true` applied *after* attach. Live remote
   queries work; residual surface is network egress via already-loaded extensions.
3. **`open`** — escape hatch, minimal restrictions, not a default.

Key sequencing rule (same pattern `PinSource` already uses — download *then* lock): **attach/materialize at construction, then lock.** The LLM never issues `ATTACH` itself.

Honest tradeoff to surface to users: a live remote attach trades the airtight `sealed` guarantee for `attached_readonly`. The LLM's SQL runs against a real (read-only) remote connection, so "the model can't touch anything external" is no longer strictly true. Snapshot mode keeps `sealed`.

### Two semantics decisions this forces

- **Result format.** A join across a pandas table and a polars pin has no single
  origin, so the session needs one output backend (e.g. `output="polars"|"pandas"`,
  default polars-if-available, matching `PinSource._convert_result`). Today each
  source returns its own library; coalescing removes that ambiguity only by choosing.
- **Freshness.** Live-attached remote = always current but every query hits the
  network; snapshot = fast but frozen at startup. Should be explicit per source
  (`materialize=True/False`), never silent — otherwise a live Postgres connection
  quietly becomes a startup full-table copy.

### Suggested rollout (additive layers)

1. **Foundation** — loosen the compat check to a `register_into` capability
   partition; generalize `DuckDBExecutor` to host any registerable source
   (dataframes + pins + files). No lockdown change. Mostly a refactor of code
   that already exists; risk-free.
2. **Remote snapshot** — materialize PG/MySQL/SQLite/Ibis into DuckDB at startup
   via ATTACH-and-copy or Arrow; lockdown stays `sealed`.
3. **Remote live** — introduce `attached_readonly`; live-attach PG/MySQL/SQLite;
   Snowflake/Databricks remain snapshot-only. Gate live mode behind an explicit
   opt-in so the lockdown relaxation is never silent.

Recommendation: build (1) now since it's a near-free refactor, structure the `register_into` contract + lockdown profiles so (2)/(3) are purely additive.

---

## 2. Better affordances for registering many tables from a database source

### Problem

Connecting an existing multi-table database is fully manual today. For a SQLAlchemy engine you add one table at a time, reusing the same engine instance:

```python
engine = sqlalchemy.create_engine("postgresql://...")
qc = QueryChat(engine, "orders")
qc.add_table(engine, "customers")   # same engine instance required
qc.add_table(engine, "products")
```

`SQLAlchemySource.__init__` already validates each table via `inspect(engine).has_table(...)` (`_datasource.py:548-550`), and the inspector exposes `get_table_names()` — but nothing calls it. For a database with 30 tables you hand-write 30 `add_table` calls. The only shipped DB example (`pkg-r/inst/examples-shiny/sqlite/app.R`) is single-table; there is no multi-table-from-a-database example in either package.

### Suggested affordances

- **Auto-discovery on construction (preferred)** — let `QueryChat(engine)` with no
  table name register every table in the (default schema of the) database, via a
  constructor argument or a dedicated method. This is the affordance to prioritize:
  it's the natural "point me at the database" path. Users who want a curated subset
  already have `.add_table()`; we don't need a separate bulk helper that only works
  in some cases. Consider an optional limit / allowlist so a 200-table warehouse
  doesn't bloat the system prompt — but the default should be "discover everything."
- **A multi-table database example** in both packages, to document the
  same-engine-instance requirement and the auto-discovery API.

---

## 3. Simplification opportunities within the PR as-is

### A. `StateDictAccessorMixin` is structurally untyped — 15 `# type: ignore[attr-defined]`

`_querychat_core.py:89` defines `StateDictAccessorMixin` with the state-dict `df()/sql()/title()` accessors, but it reaches into `self._data_sources`, `self._query_executor`, `self.client`, `self.greeting`, `self._require_query_executor`, `self._require_initialized` — none of which it owns. Result: **15 `# type: ignore[attr-defined]`** in one file.

The mixin is only mixed into Dash (`_dash.py:39`) and Gradio (`_gradio.py:37`).

**Recommendation:** introduce an intermediate base `class StateDictQueryChat(QueryChatBase)` that holds these accessors; Dash and Gradio extend *that* instead of mixing in. The methods then typecheck naturally (real class with the real attributes), all 15 ignores disappear, and Shiny / Streamlit never inherit accessors that don't fit them. (A `Protocol`-bound mixin is a viable alternative but duplicates the attribute list; the intermediate base is cleaner.)

**Why not the simpler "fold into `QueryChatBase`"?** It looks tempting but breaks, because the accessor signatures are framework-divergent —

- Dash/Gradio (mixin): `df(self, state, *, table=None)` — takes serialized state
- Streamlit (`_streamlit.py:325`): `df(self) -> IntoFrameT` — no state arg
- Shiny (`_shiny.py:896, 917`): `df(self)`, and `sql(self, query=None) -> str | None | bool`
  is an overloaded getter/setter

Hoisting the `state`-taking signature onto the root base would force Streamlit and Shiny to override with incompatible signatures — a Liskov/typing mess worse than the ignores. The mixin's *existence* is justified; only its typing is wrong. The intermediate base keeps the methods with exactly the two frameworks that share the signature.

### B. The "block the old single-table API" rule is encoded twice, with copy-pasted messages

Backward compat keeps the flat `qc_vals.df()/.sql()/.title()` accessors. In multi-table mode they must redirect to `.table('name')`. That one rule is implemented in two unrelated mechanisms with duplicated message text:

- State-dict path: three near-identical inline blocks in
  `_querychat_core.py:131-136, 181-186, 215-219`
- Shiny reactive path: a sentinel class `_MultiTableBlockedReactive`
  (`_shiny_module.py:91-107`), installed at `:385-394`

Good news — the flat accessor is genuinely cheap to keep, because in single-table mode it already just delegates to the one table's per-table state (`_shiny_module.py:371-381`: `df=only_state.df`, etc.). There is no parallel single-table path to maintain; the scaffolding exists *only* for the multi-table case.

**We do want to preserve the upgrade path** — existing single-table apps must not break, and the flat API should stay a first-class supported alias for the single-table case (not a deprecated shim). The break only matters when a user opts into a second table, which is a new action and a natural teaching moment.

**Recommendation — soft landing instead of a hard raise:**

- **Single-table:** flat accessor remains a permanent supported alias (unchanged).
- **Multi-table:** the flat accessor **delegates to the primary (first-registered)
  table** and emits a **`FutureWarning`** naming the table and pointing to
  `.table('name')`. No break, even for apps that grow into multi-table.
  - Default to the *primary* table specifically because, for an upgrading app, it
    is the table the existing `.df()` already referred to (the one passed to the
    constructor) — so behavior is continuity-preserving, not an arbitrary guess.
  - Use `FutureWarning`, **not** `DeprecationWarning`: the latter is hidden by
    default outside `__main__`/tests, so in a deployed app the developer would
    never see it (effectively silent). `FutureWarning` shows by default.
- **One message source:** the warning text and the eventual hard-error text come
  from a single builder, replacing the ~5 copy-pasted message sites.
- **Document the endpoint:** state the version where the warning flips to an error,
  so there's a real migration window and the warning isn't perpetual.

Implementation caveats for the implementer:

- **Warn once, not per reactive read.** `.df()` re-runs on every invalidation; a
  naive `warnings.warn` would spam logs. Guard it to fire once per session/location.
- This turns `_MultiTableBlockedReactive` into a *warn-then-return-primary* proxy
  rather than a hard-raising sentinel. Comparable code, better UX; the hard-error
  sentinel only returns at the documented deprecation endpoint.

### C. Three executors where the split is type-based and partly duplicated

`_query_executor.py` has `DuckDBExecutor`, `PolarsSQLExecutor`, and `DataSourceExecutor`. The abstraction is genuinely needed for multi-table SQL (a shared connection must host all tables), but the three-way split is heavier than needed:

- `DuckDBExecutor` and `PolarsSQLExecutor` share ~19 byte-identical lines of
  `test_query` / `require_all_columns` validation, differing only in how column
  names are read (`result.description` vs `collect_schema().keys()`), plus a
  duplicated `_table_columns` cache.
- `DataSourceExecutor` is a pure passthrough; for single-table it adds one
  indirection layer (`QueryChat.method()` → `_require_query_executor()` →
  `DataSourceExecutor` → `DataSource.method()`) that buys nothing.

**Recommendation — extract now, deepen later:**

- **Now (in this PR):** extract the shared validation into the abstract base as
  `QueryExecutor._validate_missing_columns(...)` (the base already hosts shared
  `get_schema`/`get_column_details`). `DuckDBExecutor`/`PolarsSQLExecutor` call the
  helper. Low-risk, immediate de-duplication.
- **Follow-up (tracked issue):** replace the *type-based* dispatch
  (`type(new) is not type(first)`, `_query_executor.py:270`) with a
  *capability-based* one — a single `DuckDBExecutor` that hosts anything
  implementing a `register_into(conn)` contract, with `DataSourceExecutor`
  demoted to a native-engine fallback (SQLAlchemy/Ibis). This removes the
  DuckDB/Polars split *and* unlocks the blended-source coalescing from section 1 —
  one refactor pays three ways.

### D. System-prompt rebuilds — keep the staged pattern, but guard prompt caching

The `add_table` staged-rebuild (`_querychat_base.py:396-418`) — build the new system prompt and executor against a *copy* of the sources dict, committing only if both succeed — is a sound transactional pattern that prevents partial-state corruption on a failed add. Keep it.

But one thing occurred to us while reading it: **rebuilding the system prompt should emit a warning if a client already exists with non-empty chat history.** Changing the system prompt changes the cached prefix, so any prompt caching established over the prior turns is invalidated — the next request silently pays full input-token cost again. If a prompt rebuild happens after a conversation has started, the developer should be told they're discarding the cache (and ideally steered toward configuring all tables *before* chatting begins). Worth a `warning` at the rebuild site whenever the bound client already has history.

cpsievert (commented):
Thanks for the detailed review. I've addressed the near-term items in sections 2 and 3; thoughts on section 1 below.

**Sections 3A–3D**

All four simplification items landed:
- [905a8db](https://github.com/posit-dev/querychat/commit/905a8db) — replaces `StateDictAccessorMixin` with `StateDictQueryChat`, eliminating all 15 `type: ignore[attr-defined]` comments
- [53a7788](https://github.com/posit-dev/querychat/commit/53a7788) — softens the flat-accessor error to `FutureWarning` with primary-table fallback, with session-scoped deduplication so Shiny reactives don't spam
- [f36db45](https://github.com/posit-dev/querychat/commit/f36db45) — extracts `_validate_missing_columns` to the `QueryExecutor` base, removing the ~19-line duplication
- [47d8fae](https://github.com/posit-dev/querychat/commit/47d8fae) — warns when `add_table` rebuilds the system prompt after a client already has chat history, covering both the `_client_spec` and `_client_console` paths

**Section 2 — Bulk table registration**

[99e011d](https://github.com/posit-dev/querychat/commit/99e011d) adds `add_tables()` for bulk registration in both R and Python, which handles the many-`add_table`-calls problem. Auto-discovery on construction is a natural next step I'd consider separately.

**Section 1 — Coalescing source types into shared DuckDB**

I've filed [#256](https://github.com/posit-dev/querychat/issues/256) to track loosening the class-identity check so `PinSource` and `DataFrameSource` can share a `DuckDBExecutor` — both already materialize an eager frame, so this is largely a routing change.

For steps 2 and 3 I'm more cautious:

*Snapshot mode* is hard to recommend as a general solution. Materializing a full remote table into local DuckDB memory at startup works for small tables and silently breaks for production ones, with no reliable way to gate on table size upfront.

*Live attach* is more interesting but comes with real costs: DuckDB's SQL dialect diverges from Postgres/MySQL enough to introduce compatibility risk (the LLM is already told what flavor to use, and "DuckDB mediating Postgres" isn't the same as talking to Postgres directly); ATTACH scanners only cover Postgres, MySQL, and SQLite, so Snowflake/Databricks users still fall back to snapshots; per-dialect connection-string translation from SQLAlchemy URLs is non-trivial glue; and relaxing `enable_external_access` is a meaningful security posture change that warrants its own rollout.

I'd keep the `register_into` contract extensible so steps 2/3 remain additive, but I'd want concrete user demand before pursuing remote-source coalescing.

cpsievert:
@gadenbuie thanks for the review, ok to merge?

gadenbuie (approved):

PR #196, opened 2026-01-15
feat(pkg-py): return native DataFrame types with generic type narrowing

cpsievert:
This PR refactors the data source classes to return native DataFrame types that match the input, rather than always wrapping results in `narwhals.DataFrame`/`narwhals.LazyFrame`. That is, if you put pandas/polars in, you (once again) get pandas/polars out. This effectively reverts the breaking change introduced in 0.4.0 and provides a more intuitive API.

### Breaking Change

This is technically a breaking change that **reverts the 0.4.0 breaking change**. Users who adapted to 0.4.0 by working with `narwhals.DataFrame` will now receive native types again:

| Input type | Return type (0.3.x) | Return type (0.4.0) | Return type (this PR) |
|------------|---------------------|---------------------|-----------------------|
| `polars.DataFrame` | `polars.DataFrame` | `narwhals.DataFrame` | `polars.DataFrame` |
| `pandas.DataFrame` | `pandas.DataFrame` | `narwhals.DataFrame` | `pandas.DataFrame` |
| `pyarrow.Table` | N/A | N/A | `pyarrow.Table` |
| `sqlalchemy.Engine` | `pandas.DataFrame` | `narwhals.DataFrame` | `narwhals.DataFrame` |

### Changes

- `DataSource` is now generic (`DataSource[T]`) with a configurable return type
- `DataFrameSource` tracks the input DataFrame library and returns results in the same format
- `PolarsLazySource` returns native `pl.LazyFrame` instead of `nw.LazyFrame`
- `SQLAlchemySource` continues to return `nw.DataFrame` (no clear "native" type to preserve)
- All `QueryChat` classes are now generic with `@overload` signatures for proper type narrowing
- Added pyarrow support for `DataFrameSource`
- Removed `duckdb_result_to_nw()` helper (no longer needed)

### Motivation

The 0.4.0 change forced users to call `.to_native()` on every result, adding friction. Returning native types:
1. Is more intuitive (you get back what you put in)
2. Enables better type inference in IDEs
3. Reduces boilerplate for downstream code

### DIFF
--- a/pkg-py/tests/test_dataframe_source.py
+++ b/pkg-py/tests/test_dataframe_source.py
@@ -3,11 +3,10 @@
 import duckdb
 import narwhals.stable.v1 as nw
 import pandas as pd
-import pytest
-from querychat._datasource import DataFrameSource
-
 import polars as pl
 import pyarrow as pa
+import pytest
+from querychat._datasource import DataFrameSource
 
 
 @pytest.fixture
--- a/pkg-py/tests/test_df_compat.py
+++ b/pkg-py/tests/test_df_compat.py
@@ -5,10 +5,9 @@ import tempfile
 from pathlib import Path
 
 import narwhals.stable.v1 as nw
+import polars as pl
 import pytest
 from querychat._df_compat import read_csv
-
-import polars as pl
 
 
 class TestReadCsv:


pkg-py/src/querychat/_datasource.py
```diff
@@ -2,22 +2,39 @@
...
+    read_sql,
+)
 from ._utils import check_query
 
 if TYPE_CHECKING:
     import polars as pl
+    from narwhals.stable.v1.typing import IntoDataFrame
     from sqlalchemy.engine import Connection, Engine
 
+# Type alias for DataFrame or LazyFrame return types
```
cpsievert:
```suggestion
```

pkg-py/src/querychat/_datasource.py
```diff
@@ -178,6 +196,10 @@ def __init__(self, df: nw.DataFrame, table_name: str):
         self._df = df
         self.table_name = table_name
 
+        # Track the native backend for returning results in the same format
+        native_namespace = nw.get_native_namespace(df)
+        self._backend = native_namespace.__name__  # "polars" or "pandas"
```
cpsievert: This feels more like `self._df_lib`

pkg-py/src/querychat/_datasource.py
```diff
@@ -285,7 +307,11 @@ def execute_query(self, query: str) -> nw.DataFrame:
 
         """
         check_query(query)
-        return duckdb_result_to_nw(self._conn.execute(query))
+        result = self._conn.execute(query)
+        if self._backend == "polars":
+            return duckdb_result_to_polars(result)
+        else:
+            return duckdb_result_to_pandas(result)
```
cpsievert: Is it possible/worth it for us to support more df libs?

cpsievert:
**Re: IntoLazyFrameT suggestion (line 94 in _dash.py)**

narwhals does have `IntoLazyFrame`, but currently Polars is the only lazy backend we support (via `PolarsLazySource`). I could add `IntoLazyFrameT` for future-proofing, but it would only match `pl.LazyFrame` today. Want me to add it now or defer until we have more lazy backends?

---

**Re: Why the `df()` method override and type ignores (_dash.py line 184)**

The override is needed to narrow the return type from `DataOrLazyFrame` (the base class return) to `DataFrameT` (the generic type parameter). The `type: ignore[override]` is because the return type doesn't match the base class signature exactly - it's more specific. I can look into making the base class generic to avoid the ignores if preferred.

---

**Re: `type: ignore[override]` on execute_query (_datasource.py line 287)**

Same reason - the base class `DataSource.execute_query()` returns `DataOrLazyFrame` (narwhals types), but the subclass returns native types (pd.DataFrame/pl.DataFrame). The override annotation suppresses the LSP violation warning.

---

**Re: Supporting more df libs (_datasource.py line 314)**

DuckDB's result conversion only has built-in methods for `.pl()` (polars) and `.df()` (pandas). Other libs would need conversion through Arrow (`.arrow()`) or similar. We could add support via Arrow as a fallback, but it adds complexity. Worth doing if there's demand for other backends like cuDF or modin.

pkg-py/src/querychat/_dash.py
```diff
@@ -86,9 +88,57 @@ def update_sql(state):
 
     """
 
+    @overload
+    def __init__(
+        self: QueryChat[pl.LazyFrame],
+        data_source: pl.LazyFrame,
```
cpsievert: It feels worth making this `IntoLazyFrameT`

pkg-py/src/querychat/_dash.py
```diff
@@ -86,9 +87,57 @@ def update_sql(state):
...
+        categorical_threshold: int = 20,
+        extra_instructions: Optional[str | PathType] = None,
+        prompt_template: Optional[str | PathType] = None,
+        storage_type: Literal["memory", "session", "local"] = "memory",
+    ) -> None: ...
+
     def __init__(
         self,
-        data_source: IntoFrame | sqlalchemy.Engine,
+        data_source: IntoDataFrame | IntoLazyFrame | sqlalchemy.Engine,
```
cpsievert: I think it makes sense to simply to this?

```suggestion
        data_source: IntoFrame | sqlalchemy.Engine,
```

pkg-py/src/querychat/_querychat_base.py
```diff
@@ -47,6 +48,9 @@ class QueryChatBase:
     - Shared methods (client, console, generate_greeting, cleanup)
 
     Framework-specific subclasses add their own UI methods.
+
+    Note: Type-parameterized overloads are defined in subclasses, not here.
+    This allows subclasses to call super().__init__() without type errors.
```
cpsievert:
```suggestion
```

pkg-py/src/querychat/_querychat_base.py
```diff
@@ -36,7 +37,7 @@
 TOOL_GROUPS = Literal["update", "query"]
 
 
-class QueryChatBase:
+class QueryChatBase(Generic[IntoFrameT]):
```
cpsievert, line 40: Does the generic actually do anything here?

cpsievert: Re: Generic[IntoFrameT] on QueryChatBase - Yes, it's needed because subclasses inherit with QueryChatBase[IntoFrameT] to parameterize the type. Without the generic on the base class, pyright errors with "Expected no type arguments for class QueryChatBase".

cpsievert:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #201, opened 2026-01-21
feat(pkg-py): add ggsql visualization tool

cpsievert:
## Summary

Adds an opt-in `visualize` tool to querychat's Python package, enabling LLM-generated data visualizations via [ggsql](https://github.com/posit-dev/ggsql). When enabled, the LLM can write ggsql queries (SQL + `VISUALISE` clause) that are executed and rendered as interactive Altair charts inline in the chat.

### Install

```bash
pip install querychat[viz]
```

This installs the optional dependencies: `ggsql`, `altair`, `shinywidgets`, and `vl-convert-python`.

### Example app

```python
from querychat.data import titanic

qc = QueryChat(
    titanic(),
    "titanic",
    tools=("query", "visualize"),
)

app = qc.app()
```

The `visualize` tool is opt-in — include it in the `tools` tuple alongside `"query"` and/or `"update"` to enable it.

### What it looks like

Charts render inline in the chat with a collapsible footer showing the ggsql query (with syntax highlighting), a copy button, and save options:

Screenshot 2026-03-20 at 7 04 42 PM

---

## Key changes

- **`visualize` tool** (opt-in via `tools=("query", "visualize")`): The LLM writes a full ggsql query, which is executed against the data source and rendered as an Altair chart.
- **Two-stage ggsql pipeline** (`_viz_ggsql.py`): Uses DataSource for SQL execution (preserving database pushdown), then feeds results into ggsql for VISUALISE processing → Altair chart.
- **Inline chart rendering** (`_viz_altair_widget.py`): Custom `AltairWidget` for rendering charts directly in the chat stream, with JS/CSS assets for interactive display.
- **Prompt restructuring**: System and tool prompts were reorganized into a unified structure. The LLM is guided to favor visualization over redundant table output, and preparatory queries are collapsed. A comprehensive ggsql syntax reference is included for the LLM.
- **Viz state is internal only**: No public accessor methods or reactive values for viz state — it's managed entirely by the module/app internals.
  - This is somewhat intentional for now given that there will likely be a "second pass" at ggsql support. At that point, Maybe it will become more clear what sort of state accessor methods might be useful beyond "give me the latest chart". 
- **Documentation**: Added a visualization guide (`docs/visualize.qmd`) with examples and screenshots.


pyproject.toml
```diff
@@ -23,12 +23,15 @@ dependencies = [
...
     "shiny>=1.5.1",
     "shinychat>=0.2.8",
+    "shinywidgets>=0.3.0",
     "htmltools",
     "chatlas>=0.13.2",
     "narwhals",
     "chevron",
     "sqlalchemy>=2.0.0", # Using 2.0+ for improved type hints and API
     "great-tables>=0.16.0",
+    "ggsql @ {root:uri}/../ggsql/ggsql-python",
```
cpsievert:
```suggestion
    "ggsql",
```

pkg-py/src/querychat/_shiny.py
```diff
@@ -870,3 +928,39 @@ def title(self, value: Optional[str] = None) -> str | None | bool:
...
+        """
+        Get the visualization title from the most recent visualize_query call.
+
+        Returns
+        -------
+        :
+            The title, or None if no visualization exists.
+
+        """
+        return self._vals.viz_title.get()
```
cpsievert: I think we should probably defer on providing a developer-facing API. It feels like there's something much more interesting here in terms of reactively reading an _interesting set_ of visuals, rather than the current visual.

cpsievert: Addressed in https://github.com/posit-dev/querychat/pull/201/changes/f3953bffbf3dd7a5af2fc2163171068d03c0e83b

pyproject.toml
```diff
@@ -21,8 +21,8 @@ maintainers = [
 ]
 dependencies = [
     "duckdb",
-    "shiny>=1.5.1",
-    "shinychat>=0.2.8",
+    "shiny @ git+https://github.com/posit-dev/py-shiny.git@feat/ggsql-language",
+    "shinychat @ git+https://github.com/posit-dev/shinychat.git@feat/react-migration",
```
cpsievert:
```suggestion
    "shinychat @ git+https://github.com/posit-dev/shinychat.git",
```

pkg-py/docs/visualize.qmd
```diff
@@ -0,0 +1,104 @@
...
+## Choosing tools
+
+The `tools` parameter controls which capabilities the LLM has access to.
+By default, only `"query"` and `"update"` are enabled — visualization must be opted into explicitly.
+
+To enable only query and visualization (no dashboard filtering):
+
+```{.python}
+qc = QueryChat(titanic(), "titanic", tools=("query", "visualize_query"))
+```
```
gadenbuie, line 42: I think `visualize` is a better tool name than `visualize_query` (which sounds like it might provide a visual representation _of a query_).

And while I'm looking at it, it's worth mentioning that I regret "update" as the tool name. It makes sense as the tool name seen by the LLM (because that's the action the LLM is taking), but in user-facing documentation we always end up explaining it as "filtering".

pkg-py/examples/10-viz-app.py
```diff
@@ -0,0 +1,26 @@
...
+#    return ui.page_fillable(
+#        qc.ui(),
+#    )
+#
+#
+#def server(input, output, session):
+#    qc.server(enable_bookmarking=True)
+#
+#
+#app = App(app_ui, server, bookmark_store="url")
```
gadenbuie: Is this left here intentionally?

pkg-py/src/querychat/prompts/prompt.md
```diff
@@ -1,4 +1,4 @@
-You are a data dashboard chatbot that operates in a sidebar interface. Your role is to help users interact with their data through filtering, sorting, and answering questions.
+You are a data dashboard chatbot that operates in a sidebar interface. Your role is to help users interact with their data through filtering, sorting, answering questions, and exploring data visually.
```
gadenbuie: This might lead to odd behavior if the visualize tool is turned off. The previous prompt was vague enough to work with at least one of the query/update tools, but it's probably worth gating the "exploring data visually" behind the presence of the visualize tool.

pkg-py/src/querychat/prompts/prompt.md
```diff
@@ -118,11 +118,105 @@ Response: "The average revenue is $X."
...
+
+#### Graceful recovery
+
+If a visualization fails, read the error message carefully and retry with a corrected query. Common fixes: correcting column names, adding `SCALE DISCRETE` for integer categories, using single quotes for strings, moving SQL expressions out of VISUALISE into the SELECT clause.{{#has_tool_query}} If the error persists, fall back to `querychat_query` for a tabular answer.{{/has_tool_query}}
+
+#### ggsql syntax reference
+
+The syntax reference below covers all available clauses, geom types, scales, and examples.
+
+{{> ggsql-syntax}}
```
gadenbuie: Since we're injecting the full prompt, I think it's worth wrapping in XML tags to help the model treat it as a subdocument. The leading sentence is just filler and can be dropped.

```suggestion
<ggqsl-syntax-reference>
{{> ggsql-syntax}}
</ggsql-syntax-reference>
```

gadenbuie: Another idea: this prompt is pretty big (approx 5.5-6k tokens) and it might be better to give the model a way to dynamically pull in the syntax guidelines when they're needed rather than including them in the system prompt. More than the raw token count, there's a reasonable chance that the additional instructions in the system prompt may make it harder for the model to follow querychat's prescribed patterns.

One way to do this would be to have the tool return the prompt, e.g on `visualize("HELP")`. We'd tell the model to run the `"HELP"` query to learn about the syntax, with instructions to run it before writing any visualization queries and to only run it once per chat. That call would return the full syntax reference, keeping system prompt lean and focused.

pkg-py/src/querychat/prompts/tool-query.md
```diff
@@ -1,17 +1,4 @@
-Execute a SQL query and return the results
```
gadenbuie, line 1: I don't think we should drop these instructions.

1. These tools might be used outside of querychat, or without querychat's system prompt
2. In my experience, models tend to follow tool instructions a little more closely than the same instructions in the system prompt
3. When I refactored the system and tool prompts, I tried to leave _coordination_ instructions in the system prompt and _tool use_ instructions in the tool description. There's definitely some overlap and I'm happy to adjust the balance but I don't think we should remove these instructions entirely.
4. I have been following a convention of putting a title/short description of the tool in the first line, followed by a more complete description. The goal is that in places like btw or apps that present a listing of the registered tools, the first line is what would show up in a tooltip.

Finally a note that there are similar prompt files for the R tools. I know we're not updating the R package here, but just to call out that we unfortunately have to keep these files manually in sync.

pkg-py/src/querychat/prompts/tool-update-dashboard.md
```diff
@@ -1,10 +1,4 @@
-Filter and sort the dashboard data
```
gadenbuie, line 1: Same comment as above

pkg-py/src/querychat/prompts/tool-visualize.md
```diff
@@ -0,0 +1,13 @@
+Render a ggsql query inline in the chat. All data transformations must happen in the SELECT clause — VISUALISE and MAPPING accept column names only, not SQL expressions or functions.
```
gadenbuie: Relatedly, I think you should try to move as much of the visualize tool-specific instructions from the system prompt into the tool description.

pkg-py/src/querychat/static/js/viz-preload.js
```diff
@@ -0,0 +1,50 @@
...
+        mutation.addedNodes.forEach(processVizPreloads);
+      });
+    });
+
+    preloadObserver.observe(document.documentElement, {
+      childList: true,
+      subtree: true,
+    });
+  }
+})();
```
gadenbuie: This was a confusing file to read, in part because there aren't any orienting comments. But also because the use of `.forEach()` implies a per-element mutation; it took a while to read and understand the code to realize the goal is to show only one `.querychat-viz-preload` element AND that we don't need to remove _other_ hidden `.querychat-viz-preload` elements.

Here's a rewrite the organizes the code around the principle of revealing one element.


```suggestion
(() => {
  // In Shiny apps, reveal the first `.querychat-viz-preload` element that appears
  // and then stop watching the DOM. This is a one-time, page-level initialization:
  // if a preload element already exists at startup, reveal it immediately; otherwise
  // observe DOM mutations until one is added, then reveal it and disconnect.

  if (!window.Shiny || window.__querychatVizPreloaded) return;

  let preloadObserver;

  const stopVizPreloadObserver = () => {
    preloadObserver?.disconnect();
    preloadObserver = undefined;
  };

  const findVizPreload = (node) => {
    if (!(node instanceof Element)) return null;
    return node.matches(".querychat-viz-preload")
      ? node
      : node.querySelector(".querychat-viz-preload");
  };

  const revealVizPreload = (root) => {
    if (!root?.isConnected || window.__querychatVizPreloaded) return false;

    window.__querychatVizPreloaded = true;
    root.hidden = false;
    stopVizPreloadObserver();
    return true;
  };

  const processVizPreloads = (node) => {
    const preloadRoot = findVizPreload(node);
    if (!preloadRoot) return false;
    return revealVizPreload(preloadRoot);
  };

  if (processVizPreloads(document.documentElement)) return;

  preloadObserver = new MutationObserver((mutations) => {
    for (const mutation of mutations) {
      for (const node of mutation.addedNodes) {
        if (processVizPreloads(node)) return;
      }
    }
  });

  preloadObserver.observe(document.documentElement, {
    childList: true,
    subtree: true,
  });
})();
```

gadenbuie (commented):
Looking great! I have a few high-level comments at the design and integration layers; I did not review the Python code in detail but I scanned the changes and nothing jumped out at me in that area.

pkg-py/docs/visualize.qmd
cpsievert: Done in b1ad65c

pkg-py/examples/10-viz-app.py
cpsievert: Good catch. I had been using this for testing and forgot to make it more user facing. I've just kept it basic for now and using Shiny Express.

pkg-py/src/querychat/prompts/prompt.md
cpsievert: Thanks, I took the XML tag suggestion. 

> dynamically pull in the syntax guidelines 

That's interesting, but it also does feel a bit like a premature optimization. The ggsql syntax only comes in when the visualize tool is explicitly added. The chances of the tool being added, but the end user not making use of it seems rather small to me. 

> may make it harder for the model to follow querychat's prescribed patterns

If we see/gather real evidence of this, I'm happy to consider it as an improvement.

pkg-py/src/querychat/prompts/prompt.md
cpsievert: Addressed in 0bec143 — the opening sentence now only mentions visual exploration when `has_tool_visualize` is true.

pkg-py/src/querychat/prompts/tool-query.md
cpsievert: Done in 231e3ef.

pkg-py/src/querychat/prompts/tool-update-dashboard.md
cpsievert: Same commit: 231e3ef

pkg-py/src/querychat/prompts/tool-visualize.md
cpsievert: Done in d4b9d20.

PR #216, opened 2026-03-27
feat(docs): add Plausible analytics to Python and R doc sites

elnelson575:
## Summary

- Adds Plausible analytics script to the Python Quarto docs via `include-in-header.html` and `_quarto.yml`
- Adds Plausible analytics script to the R pkgdown docs via inline `includes: in_header` in `_pkgdown.yml`
- Follows the same patterns used by chatlas (Quarto) and ellmer (pkgdown)

### DIFF
--- a/pkg-py/docs/_quarto.yml
+++ b/pkg-py/docs/_quarto.yml
@@ -65,6 +65,8 @@ website:
 
 format:
   html:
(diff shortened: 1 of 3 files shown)


cpsievert:
Thanks!

PR #224, opened 2026-04-28
feat(pkg-r): add ggsql visualization tool

cpsievert:
## Summary

R port of #201. Adds an opt-in `visualize` tool to querychat's R package, enabling LLM-generated data visualizations via [ggsql](https://github.com/posit-dev/ggsql). When enabled, the LLM can write ggsql queries (SQL + `VISUALISE` clause) that are executed and rendered as interactive charts inline in the chat.

## Usage

```r
library(querychat)

qc <- QueryChat$new(
  palmerpenguins::penguins,
  tools = c("filter", "query", "visualize")
)

qc$app()
```

The `visualize` tool is opt-in — include `"visualize"` in the `tools` vector alongside `"filter"` and/or `"update"` to enable it.

### What it looks like

Charts render inline in the chat with a collapsible footer showing the ggsql query (with syntax highlighting), a copy button, and save options:

Screenshot 2026-03-20 at 7 04 42 PM

---

## Key changes

- **`visualize` tool** (opt-in via `tools = c("query", "visualize")`): The LLM writes a full ggsql query, which is executed against the data source and rendered using ggsql's native Shiny bindings (`ggsql::renderGgsql` / `ggsql::ggsqlOutput`).
- **Two-stage ggsql pipeline** (`querychat_viz.R`): Uses DataSource for SQL execution (preserving database pushdown), then feeds results into ggsql's DuckDB reader for VISUALISE processing.
- **Inline chart rendering**: Charts render directly in the chat stream via Shiny's dynamic output system, with JS/CSS assets for the footer (show query toggle, save as PNG/SVG).
- **Generated footer assets**: `pkg-r/inst/htmldep/viz.{js,css}` are checked-in package assets, but the shared source/build logic for them already lives on `main` under `js/` (see `js/build.mjs`). This PR is adding the R package's generated copies, not introducing a second implementation of that browser logic.
- **PNG feedback to LLM**: A static PNG is generated (best-effort, requires `V8` + `rsvg`) and sent back to the LLM so it can see what it produced.
- **`collapsed` parameter on `querychat_query`**: Preparatory queries (e.g., inspecting data before visualizing) can start collapsed so the chart remains the focal point.
- **Prompt restructuring**: System and tool prompts were updated to match the Python package — visualization best practices, ggsql syntax reference, and conditional sections based on which tools are enabled.
- **Bookmark support**: Viz widget state is saved/restored during Shiny bookmarking.
- **Viz state is internal only**: No public accessor methods or reactive values for viz state — consistent with the Python implementation.

## Test plan

- [x] Unit tests for `execute_ggsql`, `extract_visualise_table`, `has_layer_level_source`
- [x] Unit tests for `tool_visualize_dashboard` (creation, callback, error handling)
- [x] Unit tests for `truncate_error`
- [x] Unit tests for `collapsed` parameter on query tool
- [x] Unit tests for viz prompt conditionals in system prompt
- [x] Full test suite passes (622 tests, 0 failures)


pkg-r/R/querychat_viz.R
```diff
@@ -0,0 +1,362 @@
...
+  validated <- ggsql::ggsql_validate(ggsql_str)
+  has_visual <- ggsql::ggsql_has_visual(validated)
+
+  if (!has_visual) {
+    has_keyword <- grepl("VISUALIS[EZ]", ggsql_str, ignore.case = TRUE)
+    if (has_keyword) {
+      rlang::abort(
+        "VISUALISE clause was not recognized. VISUALISE and MAPPING accept column names only — no SQL expressions, CAST(), or functions. Move all data transformations to the SELECT clause, then reference the resulting column by name in VISUALISE."
+      )
+    }
```
cpsievert: This may not be necessary anymore https://github.com/posit-dev/ggsql/pull/389

pkg-r/inst/prompts/ggsql-syntax.md
```diff
@@ -0,0 +1,561 @@
...
+   SELECT ROOM_TYPE, COUNT(*) AS listings FROM airbnb
+   VISUALISE ROOM_TYPE AS x, listings AS y
+   DRAW bar
+
+   -- CORRECT: Alias to lowercase, then reference the alias
+   SELECT ROOM_TYPE AS room_type, COUNT(*) AS listings FROM airbnb
+   VISUALISE room_type AS x, listings AS y
+   DRAW bar
+   ```
+   As a general rule, always use lowercase column names and aliases in both SELECT and VISUALISE clauses.
```
cpsievert: May not be necessary after https://github.com/posit-dev/ggsql/pull/374

cpsievert: Addressed in [afbbfba](https://github.com/posit-dev/querychat/pull/224/commits/afbbfba65c0fc518d2471ded0dddb114d3811122)

pkg-r/R/QueryChat.R
```diff
@@ -169,6 +171,21 @@ QueryChat <- R6::R6Class(
         chat$register_tool(tool_query(private$.data_source))
       }
 
+      if ("visualize" %in% tools) {
+        rlang::check_installed(
+          "ggsql",
+          reason = "for visualization support. Install with: pak::pak('posit-dev/ggsql-r')"
```
cpsievert:
```suggestion
          reason = "for visualization support."
```

pkg-py/src/querychat/_viz_tools.py
```diff
@@ -313,3 +324,8 @@ def build_viz_footer(
     )
 
     return TagList(buttons_row, query_section)
+
+
+def resolve_viz_dom_id(widget_id: str) -> str:
+    """Resolve a widget id to the final DOM id rendered by Shiny."""
+    return str(resolve_id(widget_id))
```
cpsievert: Get rid of this helper.

pkg-py/src/querychat/_viz_tools.py
```diff
@@ -281,10 +289,11 @@ def build_viz_footer(
...
                 {"class": "querychat-save-dropdown"},
                 tags.button(
-                    {
-                        "class": "querychat-save-btn",
-                        "data-widget-id": widget_id,
-                    },
+                        {
+                            "class": "querychat-save-btn",
+                            "data-querychat-action": "save-toggle",
+                            "data-widget-id": dom_widget_id or widget_id,
```
cpsievert: Do we really need the `or` fallback?

cpsievert: Dropped the fallback. `build_viz_footer()` now requires the resolved DOM id and writes it directly into `data-widget-id`.

pkg-py/src/querychat/_viz_tools.py
cpsievert: Removed the helper and inlined `str(resolve_id(widget_id))` at the call site in `VisualizeResult`.

pkg-r/R/querychat_viz.R
```diff
@@ -0,0 +1,383 @@
...
+            `data-widget-id` = dom_widget_id,
+            `data-title` = title,
+            "Save as PNG"
+          ),
+          shiny::tags$button(
+            class = "querychat-save-svg-btn",
+            `data-querychat-action` = "save-svg",
+            `data-widget-id` = dom_widget_id,
+            `data-title` = title,
+            "Save as SVG"
```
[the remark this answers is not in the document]
cpsievert: Not worth addressing — the title ends up in `link.download` which browsers already sanitize (strip path separators, truncate, etc.). Worst case is an ugly filename, not a security issue, and LLM-generated titles are short descriptive strings in practice.

pkg-r/R/querychat_module.R
```diff
@@ -121,6 +135,9 @@ mod_server <- function(
         state$values$querychat_sql <- current_query()
         state$values$querychat_title <- current_title()
         state$values$querychat_has_greeted <- has_greeted()
+        if (length(viz_widgets) > 0) {
+          state$values$querychat_viz_widgets <- viz_widgets
+        }
```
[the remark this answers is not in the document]
cpsievert: Not going to address this one — the ggsql strings are relatively short, and in practice users won't accumulate enough visualizations in a single session to blow past URL limits. If it ever becomes a real problem we can revisit, but it feels premature to add complexity here.

cpsievert:
@gadenbuie ready for your eyes if you'd like to have a look. CI is failing since `ggsql` isn't yet on CRAN, but it should happen in the next few days. Will get CI is passing before merging.

PR #246, opened 2026-06-11
feat: add PinSource data source for pins boards

gadenbuie:
## Summary

Adds a new `PinSource` data source to both the R and Python querychat packages, letting users chat with data stored as [pins](https://pins.rstudio.com/). Users can pass a `PinSource` directly or hand a pins board to `QueryChat`/`querychat_app()` and let `normalize_data_source()` create one automatically.

**Key design decisions:**

- **Two loading paths:** For parquet, CSV, and JSON pins, DuckDB reads the file directly via `pin_download()` + `CREATE TABLE AS SELECT * FROM read_parquet(...)`, skipping R/Python deserialization entirely. For other pin types (rds, arrow, etc.), falls back to `pin_read()` + in-memory registration.
- **DuckDB security lockdown** after data loading: `enable_external_access = false`, `autoinstall_known_extensions = false`, `autoload_known_extensions = false` — prevents LLM-generated SQL from accessing the filesystem.
- **Auto-populates `data_description`** from pin metadata (title, description, tags) when the user hasn't provided their own.
- **Pins board shorthand:** `normalize_data_source()` now accepts a pins board directly as `data_source`, creating a `PinSource` automatically.
- **Table name inference:** `QueryChat` infers `table_name` from `DataSource.table_name` when not provided, and `PinSource` sanitizes pin names (e.g. `"alex.chisholm/posit_ai_interest"` → `"alex_chisholm_posit_ai_interest"`).
- **Shared DuckDB helpers (Python):** Extracted `duckdb_column_meta()`, `duckdb_column_stats()`, and `duckdb_get_schema()` as shared functions, eliminating ~90 lines of duplication between `DataFrameSource` and `PinSource`.

**Additional fixes included:**

- Old data source is now cleaned up (`cleanup()`) when replaced via the `data_source` setter, preventing DuckDB connection leaks in long-lived apps.
- Multi-file pins are rejected with a clear error message.
- `pins` is an optional dependency in both R (`Suggests`) and Python (`[project.optional-dependencies]`).

### R

```r
library(pins)
library(querychat)

board <- board_folder(withr::local_tempdir())
pin_write(board, mtcars, "mtcars", title = "Motor Trend Car Data")

# Direct PinSource usage
ps <- PinSource$new(board, "mtcars")
ps$execute_query("SELECT mpg, cyl FROM mtcars LIMIT 5")
ps$get_data_description()  # "Motor Trend Car Data"

# Shorthand: pass board directly
app <- querychat_app(board, "mtcars", greeting = "Ask me about cars!")
shiny::runApp(app)
```

### Python

```python
import pins
from querychat import PinSource
import pandas as pd

board = pins.board_folder("/tmp/test-board")
board.pin_write(pd.DataFrame({"x": [1, 2], "y": [3, 4]}), "mydata", title="Test Data", type="parquet")

ps = PinSource(board, "mydata")
ps.execute_query("SELECT * FROM mydata")
ps.get_data_description()  # "Test Data"
```


pkg-py/src/querychat/_querychat_base.py
```diff
@@ -58,7 +59,7 @@ class QueryChatBase(Generic[IntoFrameT]):
     def __init__(
         self,
         data_source: IntoFrame | sqlalchemy.Engine | None,
```
cpsievert: Any reason why we wouldn't add `BaseBoard` to supported `data_source` types?

pkg-py/src/querychat/_querychat_base.py
```diff
@@ -68,10 +69,30 @@ def __init__(
...
         self._table_name = table_name
 
-        if not re.match(r"^[a-zA-Z][a-zA-Z0-9_]*$", table_name):
+        is_pins_board = False
+        try:
+            from pins.boards import BaseBoard
+
+            is_pins_board = isinstance(data_source, BaseBoard)
+        except ImportError:
+            pass
```
cpsievert: Pull this logic out into a `TypeGuard` helper that lives in `_pin_source.py`

pkg-py/src/querychat/_querychat_base.py
cpsievert: Same question as https://github.com/posit-dev/querychat/pull/246#discussion_r3399662985

pkg-py/src/querychat/_querychat_base.py
```diff
@@ -90,14 +112,28 @@ def __init__(
...
 
+    def _auto_fill_data_description(self) -> None:
+        """Auto-populate data_description from PinSource metadata if not user-supplied."""
+        if self._data_description_auto:
+            self._data_description = None
+            self._data_description_auto = False
+        if self._data_description is None and self._data_source is not None:
+            desc = self._data_source.get_data_description()
+            if desc:
+                self._data_description = desc
```
cpsievert, line 130: Do we want the pins description to fully overwrite a user-provided description?

pkg-py/src/querychat/_querychat_base.py
```diff
@@ -90,14 +112,28 @@ def __init__(
...
+            self._table_name = self._data_source.table_name
+            self._auto_fill_data_description()
             self._build_system_prompt()
         else:
             self._data_source = None
             self._system_prompt = None
 
+    def _auto_fill_data_description(self) -> None:
+        """Auto-populate data_description from PinSource metadata if not user-supplied."""
+        if self._data_description_auto:
```
cpsievert: I feel like it might be more clear what this method is doing if the flag was instead something like `_data_description_mode: Literal["supplied", "inferred", "empty"]`?

pkg-py/src/querychat/__init__.py
```diff
@@ -1,9 +1,11 @@
 from ._deprecated import greeting, init, sidebar, system_prompt
 from ._deprecated import mod_server as server
 from ._deprecated import mod_ui as ui
+from ._pin_source import PinSource
 from ._shiny import QueryChat
 
 __all__ = (
+    "PinSource",
```
cpsievert: Let's only export this in the `types` module, especially if we're going to support passing a baseboard directly to `QueryChat()`.

cpsievert (commented):
Thanks, this is a nice addition! Here are some initial thoughts from scanning the Python logic. I haven't looked yet at the R changes.

pkg-py/src/querychat/_querychat_base.py
gadenbuie: No — a user-provided `data_description` is never overwritten. The new `_data_description_mode` tracks this explicitly:

- `"supplied"` (user provided a value) → `_auto_fill_data_description()` is a no-op, pin metadata is ignored
- `"empty"` (no description) → tries to infer from `data_source.get_data_description()`
- `"inferred"` (auto-filled) → clears on source swap, then re-infers from the new source

The `"supplied"` mode is sticky — it never transitions to `"empty"` or `"inferred"`, so user descriptions survive source swaps too.

See b5901f8 for the mode enum change and 5f3206b for a test that explicitly verifies a user-supplied description survives a data source swap.

cpsievert:
Have you considered switching `PinSource` to return narwhals DataFrames (backed by polars) instead of pandas?

- DuckDB's `.pl()` skips the Arrow→pandas conversion that `.df()` does, so every query gets faster for free. Wrapping the result in `nw.from_native()` gives consumers a uniform type.
- `SQLAlchemySource` already sets this precedent — it's `DataSource[nw.DataFrame]` because, like `PinSource`, it doesn't have a user-provided input format to mirror back.
- With polars available, we could also add `arrow`/`feather` to the direct-read set via `pl.read_ipc()` + DuckDB register. That would put all four common tabular pin types on the fast path instead of falling through to `pin_read()`.

Practically it'd mean `PinSource(DataSource[nw.DataFrame])`, `.pl()` → `nw.from_native()` instead of `.df()`, and swapping the `pins` extra to `["pins>=0.9.1", "polars", "pyarrow"]`.

gadenbuie:
> Have you considered switching `PinSource` to return narwhals DataFrames (backed by polars) instead of pandas?

Thanks for catching this; I meant to call it out in the description. I was hoping to get a vibe check from about whether that was a good idea or not. Seems like it would be!

pkg-py/src/querychat/_pin_source.py
```diff
@@ -0,0 +1,173 @@
...
+
+            conn.execute("""
+SET allow_community_extensions = false;
+SET allow_unsigned_extensions = false;
+SET autoinstall_known_extensions = false;
+SET autoload_known_extensions = false;
+SET enable_external_access = false;
+SET disabled_filesystems = 'LocalFileSystem';
+SET lock_configuration = true;
+            """)
```
cpsievert: Worth pulling this out into a helper (on both the R and Python side)?

pkg-py/src/querychat/_pin_source.py
```diff
@@ -0,0 +1,173 @@
...
+                )
+
+        return result
+
+    def get_data(self) -> pd.DataFrame:
+        return self._conn.execute(f'SELECT * FROM "{self.table_name}"').df()
+
+    def cleanup(self) -> None:
+        if self._conn:
+            self._conn.close()
```
cpsievert, line 227: Could remove this if you wanted since it's the same implementation as the base method.

pkg-r/R/QueryChat.R
```diff
@@ -111,6 +112,20 @@ QueryChat <- R6::R6Class(
       }
     },
 
+    auto_fill_data_description = function() {
+      if (private$.data_description_auto) {
```
cpsievert: Worth changing this to `.data_description_mode` for consistency with Python?

gadenbuie:
> Have you considered switching PinSource to return narwhals DataFrames (backed by polars) instead of pandas?

Okay, done in ad18fa7 and def2051.

PinSource now returns `nw.DataFrame`, consistent with SQLAlchemySource. The conversion uses a polars-if-available pattern (same as `read_sql()` in `_df_compat.py`):

```python
def _convert_result(result: duckdb.DuckDBPyConnection) -> nw.DataFrame:
    if _has_polars():
        return nw.from_native(result.pl())
    return nw.from_native(result.df())
```

Polars gives you the fast `.pl()` path but isn't required — pandas fallback works fine. I dropped the explicit `pandas` from the `pins` extra since `pins` itself hard-depends on it.

I also added arrow/IPC pin support: when polars is available, PinSource uses `pl.read_ipc()` to read arrow pins directly (DuckDB has no native IPC reader), falling back to `pin_read()` otherwise. So now the four common tabular pin types (parquet, csv, json, arrow) all have a fast path when polars is installed.

pkg-py/src/querychat/_pin_source.py
gadenbuie: Done in a48425b — extracted `duckdb_lock_down()` in both R (`utils-duckdb.R`) and Python (`_datasource.py`). Replaces inline duplicates in DataFrameSource and PinSource on both sides.

pkg-py/src/querychat/_pin_source.py
gadenbuie: Can't remove this one — PinSource inherits from `DataSource` directly (not `DataFrameSource`), and `get_db_type()` is abstract on the base class, so it must be implemented.

pkg-r/R/QueryChat.R
gadenbuie: Done in a48425b — renamed `.data_description_auto` (boolean) to `.data_description_mode` with the same 3-state values as Python: `"supplied"`, `"inferred"`, `"empty"`.

pkg-py/src/querychat/_pin_source.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def _convert_result(result: duckdb.DuckDBPyConnection) -> nw.DataFrame:
+    """Convert a DuckDB result to a narwhals DataFrame, preferring polars if available."""
+    if _has_polars():
+        return nw.from_native(result.pl())
+    return nw.from_native(result.df())
+
+
+class PinSource(DataSource[nw.DataFrame]):
```
cpsievert, line 64: I think what we have here is solid, but I also just want to point out that we could probably also have a `PinLazySource(DataSource[nw.LazyFrame])` that skips the whole download+import to DuckDB step.

Now that I'm saying this out loud though, maybe we could just document how to feed a pins URL into a `pl.LazyFrame()` and lean into the existing `PolarsLazySource` class?

cpsievert (approved):
LGTM once outstanding comments are addressed

pkg-r/R/QueryChat.R
```diff
@@ -286,7 +301,11 @@ QueryChat <- R6::R6Class(
             "{.arg table_name} is required when {.arg data_source} is {.val NULL}."
           )
         }
-        if (is.data.frame(data_source) || inherits(data_source, "tbl_sql")) {
+        if (inherits(data_source, "DataSource")) {
+          table_name <- data_source$table_name
```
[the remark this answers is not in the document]
gadenbuie: Fixed in 80a067a — added `inherits(data_source, "pins_board")` check in all three locations (`QueryChat$initialize`, `querychat()`, `querychat_app()`) that errors early with a clear message when a board is passed without a pin name.

pkg-r/R/QueryChat.R
```diff
@@ -805,10 +832,15 @@ QueryChat <- R6::R6Class(
...
         private$.data_source
       } else {
+        old_source <- private$.data_source
         private$.data_source <- normalize_data_source(
           value,
           private$.table_name
         )
+        if (!is.null(old_source)) {
+          old_source$cleanup()
+        }
```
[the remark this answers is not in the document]
gadenbuie: Fixed in 64dff4b — added `!identical(old_source, private$.data_source)` guard before calling `cleanup()`.

pkg-py/src/querychat/_querychat_base.py
```diff
@@ -237,9 +269,15 @@ def data_source(self) -> DataSource | None:
...
-    def data_source(self, value: IntoFrame | sqlalchemy.Engine) -> None:
+    def data_source(self, value: IntoFrame | sqlalchemy.Engine | BaseBoard) -> None:
         """Set the data source, normalizing and rebuilding system prompt."""
+        old_source = self._data_source
+        if self._table_name is None:
+            raise ValueError("table_name must be set before assigning a data source")
         self._data_source = normalize_data_source(value, self._table_name)
+        if old_source is not None:
+            old_source.cleanup()
+        self._auto_fill_data_description()
```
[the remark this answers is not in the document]
gadenbuie: Fixed in ec6e730 — added `old_source is not self._data_source` guard before calling `cleanup()`.

pkg-py/src/querychat/_pin_source.py
gadenbuie: Good idea. I added documentation for this pattern in e420c94e7c545116e0ab37b57126371473fc34e6 — both the data-sources vignettes (R and Python) and the PinSource class docs now have a "Lazy queries with pins" section showing how to use `pin_download()` + `tbl_sql` (R) or `pl.scan_parquet()` + `PolarsLazySource` (Python) for large parquet pins.

The docs note that `pin_download()` still fetches files to a local cache, so "lazy" here means deferred query execution, not deferred data access. They also call out that the lazy path skips the DuckDB security lockdown that PinSource applies.

A proper `PinLazySource` or upstream `pin_url()` could be interesting follow-ups if there's demand.

PR #251, opened 2026-06-18
feat(pkg-r): support sqlite engine in PinSource

gadenbuie:
## Summary

`PinSource` previously always used DuckDB, even for pins that deserialize to a plain data frame. Since `duckdb` and `RSQLite` are both optional (`Suggests`) in the R package, a user who installed only RSQLite could use `DataFrameSource` but not `PinSource`. This PR aligns `PinSource` with `DataFrameSource`'s engine handling.

Key changes:

- Extracted `DataFrameSource`'s inline connection setup into a shared internal helper, `new_dataframe_connection(df, table_name, engine)`, which creates an in-memory connection, registers the data frame, and applies the DuckDB security lockdown. `DataFrameSource$initialize()` now calls it.
- `PinSource$new()` gains an `engine` argument (default `getOption("querychat.DataFrameSource.engine", NULL)`, resolved via the same `get_default_dataframe_engine()` as `DataFrameSource`). Data-frame pins now share `DataFrameSource`'s code path, so the SQL flavor reported by the inherited `get_db_type()` is resolved consistently.
- The efficient direct DuckDB file read is kept for parquet/CSV/JSON pins under the `duckdb` engine. Requesting `engine = "sqlite"` for those pin types falls back to `pin_read()` with a warning that DuckDB reads them more efficiently.
- `duckdb` is no longer unconditionally required by `PinSource` — it is only needed when actually used.

While here, this PR also backfills a changelog entry for the original `PinSource` feature (#246), which was merged without one, in both `NEWS.md` and the Python `CHANGELOG.md`. The Python side is otherwise intentionally unchanged: `duckdb` is a core dependency there and `DataFrameSource`/`PinSource` are DuckDB-only by design, so no equivalent gap exists.

## Verification

```r
board <- pins::board_temp()
pins::pin_write(board, mtcars, "mtcars", type = "rds")

# Force the SQLite engine for a data-frame pin
ps <- querychat::PinSource$new(board, "mtcars", engine = "sqlite")
ps$get_db_type()
#> [1] "SQLite"
ps$execute_query("SELECT * FROM mtcars WHERE mpg > 30")

# A file-type pin under sqlite warns, then reads via pin_read()
pins::pin_write(board, mtcars, "cars", type = "csv")
ps2 <- querychat::PinSource$new(board, "cars", engine = "sqlite")
#> Warning: Reading csv pin "cars" into SQLite.
#> i The duckdb engine reads csv pins more efficiently. ...
ps2$cleanup()
ps$cleanup()
```

New tests cover the SQLite data-frame path and the file-type fallback warning in `tests/testthat/test-PinSource.R`.

### DIFF
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### New features
+
+* Added `PinSource`, a data source for chatting with datasets pinned to a [pins](https://pins.rstudio.com/) board. Works with parquet, CSV, JSON, and Arrow pins, and uses the pin's title, description, and tags as the default data description. Install the optional dependency with `pip install querychat[pins]`. (#246)
+
 ### Improvements
 
 * The query tool result card now starts collapsed by default. Users can still expand it to see the SQL query and results. Set `QUERYCHAT_TOOL_DETAILS=expanded` to restore the previous behavior. (#239)
--- a/pkg-r/NEWS.md
+++ b/pkg-r/NEWS.md
@@ -1,5 +1,9 @@
 # querychat (development version)
 
+## New features
+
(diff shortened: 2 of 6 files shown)


gadenbuie:
Python E2E failures are unrelated, this PR only touches R code

cpsievert (approved):

PR #252, opened 2026-06-18
test(pkg-py): fix Playwright chat input placeholder assertions

gadenbuie:
The Python E2E (Playwright) tests started failing in CI because Shiny's chat input migrated upstream to a TipTap/ProseMirror `contenteditable` div. The placeholder is now exposed via `data-placeholder` instead of the native `placeholder` attribute, so `to_have_attribute("placeholder", ...)` resolved to `None`.

Updated the three affected Shiny-based assertions to check `data-placeholder`. The Gradio test is unchanged since it targets a real `<textarea>` that still uses `placeholder`.


cpsievert:
Thanks!

PR #261, opened 2026-06-24
feat: schema-aware greetings via a QueryChatGreeter API

gadenbuie:
Supersedes the constructor-based approach in #260 (kept open for discussion). Credit to @cpsievert for the regression insight and test infrastructure.

## Summary

After multi-table support (#195), the schema moved out of the system prompt into a lazy `tool_get_schema` tool. Greetings are generated by a tool-free client, so the model became schema-blind when writing the opening message — it could no longer describe the data it was about to chat about.

This PR fixes that with a dedicated greeting API rather than a constructor parameter. The greeting-specific concern lives on a new semi-internal `QueryChatGreeter`, reached via `qc$greeter` (R) / `qc.greeter` (Python), instead of widening the `QueryChat` constructor for every single-table user.

Key design decisions:

- **`QueryChatGreeter`** holds `$tables` (table names whose schema to embed in the greeting) and `$prompt` (the greeting template). It is semi-internal — accessed through `qc.greeter`, not separately constructed (R class is `@noRd`). Setters store only; changing greeter config never invalidates an existing `qc$greeting`, so a user-supplied static greeting always survives.
- **Separate greeting system prompt** (`greeting.md`) instead of the full `prompt.md`. The greeting client no longer receives SQL guidelines or tool descriptions — just the schema for the selected tables, the data description, and suggestion-card syntax. The greeter owns greeting-client construction via `build_client()`, `generate()` (sync), and `generate_stream()` (async/Shiny).
- **`QueryChatSystemPrompt` gained granular include flags** — `include_tables`, `include_relationships`, `include_glossary` — so the greeting path can request a scoped prompt without inline filtering. The multi-table-no-dict warning fires only on the session path; greetings never re-warn.
- **Constructor tables are always included** in the greeting; `add_table(include_in_greeting=)` / `add_tables(include_in_greeting=)` opt additional tables in (default off). `add_tables` accepts a logical/`bool` or a character vector / `list[str]` of names.
- **Greeting data dicts are scoped** to the included tables: per-table entries are filtered, and table-less global dicts keep their description but drop cross-table relationships/glossary, so a curated greeting subset can't leak excluded-table metadata.
- **Greetings generate on a separate, throwaway client** in every backend (Shiny, Streamlit, Gradio, Dash). The shared `qc.greeting` is written only by an explicit `generate_greeting()` / `greeter.generate()`; in-app greetings stay session-local.
- **Client resolution is split from cloning** (Python). `resolve_client()` turns a spec into a `Chat` object (done once at init); `create_client()` cheaply clones it per session/greeting. Invalid provider strings now fail at init instead of at first session.

The Python and R implementations mirror each other. `generate_greeting()` is kept as a thin wrapper over `greeter.generate()`, so existing callers need no changes.

## Verification

Tables passed to the constructor are automatically included in the greeting. Tables added later with `add_table()` are **not** included by default — pass `include_in_greeting = TRUE` to opt them in.

R:

```r
library(querychat)
qc <- QueryChat$new(mtcars, "mtcars")
# Constructor table is included in the greeting automatically
qc$greeter$tables
#> [1] "mtcars"

# add_table() does NOT include the table in the greeting by default
qc$add_table(flights, "flights")
qc$greeter$tables
#> [1] "mtcars"

# ...opt it in explicitly
qc$add_table(airports, "airports", include_in_greeting = TRUE)
qc$greeter$tables
#> [1] "mtcars"   "airports"

greeting <- qc$generate_greeting()  # builds a lean, schema-aware greeting client
```

Python:

```python
qc = QueryChat(data_source=df, table_name="mtcars")
qc.greeter.tables          # ['mtcars'] -- constructor table auto-included

# add_table() does NOT include the table in the greeting by default
qc.add_table(flights, "flights")
qc.greeter.tables          # ['mtcars']

# ...opt it in explicitly
qc.add_table(airports, "airports", include_in_greeting=True)
qc.greeter.tables          # ['mtcars', 'airports']

# add_tables() takes a list of names to include a subset
qc.add_tables(engine, include_in_greeting=["orders", "customers"])

qc.generate_greeting()     # schema-aware greeting on a separate client
```

### DIFF
--- a/pkg-py/src/querychat/_querychat_base.py
+++ b/pkg-py/src/querychat/_querychat_base.py
@@ -341,10 +341,16 @@ class QueryChatBase(Generic[IntoFrameT]):
         """
         tbls = [n for n in self.greeter.tables if n in self._data_sources]
         sources = {n: self._data_sources[n] for n in tbls}
-        # Only keep dicts that describe an included table, so a curated greeting
-        # subset doesn't carry dict-level prose about excluded tables.
+        # Keep a dict if it describes an included table, or if it is a global
+        # (table-less) dict carrying a dict-level description. Drop the
+        # cross-table global fields (relationships, glossary) so a curated
+        # greeting subset can't leak excluded-table prose; per-table entries are
+        # scoped to the included tables at render time.
         greeting_dicts = [
-            dd for dd in self._data_dicts if any(n in tbls for n in dd.tables)
+            dd.model_copy(update={"relationships": [], "glossary": {}})
+            for dd in self._data_dicts
+            if any(n in tbls for n in dd.tables)
+            or (not dd.tables and dd.description)
         ]
         greeting_prompt_obj = QueryChatSystemPrompt(
             prompt_template=self.greeter.prompt,
--- a/pkg-py/src/querychat/_querychat_base.py
+++ b/pkg-py/src/querychat/_querychat_base.py
@@ -934,6 +934,7 @@ class StateDictQueryChat(QueryChatBase[IntoFrameT]):
             client_factory=self._client_factory,
             greeting=self.greeting,
             query_executor=self._require_query_executor("_deserialize_state"),
+            greeting_client_factory=self._build_greeting_client,
         )
         if state_data:
             state.update_from_dict(state_data)
--- a/pkg-py/src/querychat/_querychat_base.py
+++ b/pkg-py/src/querychat/_querychat_base.py
@@ -626,8 +626,7 @@ class QueryChatBase(Generic[IntoFrameT]):
         for name in greeting_names:
             if name not in new_greeting:
                 new_greeting.append(name)
-        if new_greeting != self.greeter.tables:
-            self.greeter.tables = new_greeting
+        self.greeter.tables = new_greeting
 
     def remove_table(self, table_name: str) -> None:
         """
@@ -667,6 +666,10 @@ class QueryChatBase(Generic[IntoFrameT]):
 
         self._build_system_prompt(data_sources=next_data_sources)
         self._data_sources = next_data_sources
+        if self._greeter is not None:
+            self._greeter.tables = [
+                n for n in self._greeter.tables if n != table_name
+            ]
         if self._query_executor is not None:
             with contextlib.suppress(Exception):
                 self._query_executor.cleanup()
--- a/pkg-py/src/querychat/_querychat_base.py
+++ b/pkg-py/src/querychat/_querychat_base.py
@@ -479,7 +479,7 @@ class QueryChatBase(Generic[IntoFrameT]):
         tables: list[str] | None = None,
         *,
         replace: bool = False,
-        include_in_greeting: bool | list[str] = False,
+        include_in_greeting: bool | str | list[str] = False,
     ) -> None:
         """
         Add multiple tables from a SQLAlchemy engine or Ibis backend in a single call.
@@ -502,7 +502,8 @@ class QueryChatBase(Generic[IntoFrameT]):
             name already exists.
         include_in_greeting
             ``True`` to include all added tables in the greeting, ``False`` (default)
-            for none, or a list of table names to include.
+            for none, or a table name (or list of table names) to include. Any
+            other type raises ``TypeError``.
 
         Raises
         ------
@@ -591,12 +592,22 @@ class QueryChatBase(Generic[IntoFrameT]):
                 self._query_executor.cleanup()
             self._query_executor = None
 
-        if include_in_greeting is True:
-            greeting_names = list(tables)
-        elif include_in_greeting is False:
-            greeting_names = []
+        if isinstance(include_in_greeting, bool):
+            greeting_names = list(tables) if include_in_greeting else []
+        elif isinstance(include_in_greeting, str):
+            greeting_names = (
+                [include_in_greeting] if include_in_greeting in tables else []
+            )
+        elif isinstance(include_in_greeting, list) and all(
+            isinstance(name, str) for name in include_in_greeting
+        ):
+            greeting_names = [name for name in include_in_greeting if name in tables]
         else:
-            greeting_names = [n for n in include_in_greeting if n in tables]
+            raise TypeError(
+                "include_in_greeting must be True, False, or a table name "
+                "(or list of table names), got "
+                f"{type(include_in_greeting).__name__}."
+            )
 
         new_greeting = list(self.greeter.tables)
         for name in greeting_names:
--- a/pkg-py/src/querychat/_querychat_base.py
+++ b/pkg-py/src/querychat/_querychat_base.py
@@ -422,6 +422,8 @@ class QueryChatBase(Generic[IntoFrameT]):
 
         Raises
         ------
+        TypeError
+            If include_in_greeting is not a bool.
         ValueError
             If table_name already exists (and replace=False) or is invalid.
         RuntimeError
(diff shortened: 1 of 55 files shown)


cpsievert (approved):
Thanks! If you want to update the docs/news as well, that'd be great

gadenbuie:
## Refinement: client lifecycle + greeter as hub

The original `build_greeting_client` on `QueryChat` was doing two unrelated jobs: assembling the greeting system prompt (greeting-specific policy) and materializing a client from a spec (generic lifecycle shared with the session path). This round separates those concerns and makes the greeter the single hub for all greeting logic.

### 1. Extract `create_client` primitive (af44bc2)

R gained a `create_client()` helper in `utils-ellmer.R` (resolve spec → clone → clear turns). Deduplicates the client-lifecycle dance between `create_session_client` and the greeting path. Python already had this.

### 2. Granular include flags on `QueryChatSystemPrompt` (5aa9640, f8fc4f0)

New constructor params: `include_tables` (logical/char or bool/list[str]), `include_relationships`, `include_glossary`. Greeting table-subset filtering and dict keep/strip logic moved **into** the prompt class (was inline in the old `build_greeting_client`). Multi-table-no-dict warning now fires only on the session path (greeting never re-warns).

### 3. Greeter is the hub (480a4f7, 6635f24)

`QueryChatGreeter` now owns greeting-client construction. Public surface:

- `build_client(base=NULL)` — fresh greeting client (the shared primitive)
- `generate(echo)` — sync, returns text (console path)
- `generate_stream(...)` — async/Shiny: build client → stream into chat UI → capture into session-local reactive

`build_greeting_client` removed from `QueryChat`/`QueryChatBase`. `generate_greeting(echo)` delegates to `greeter$generate(echo)`. `mod_server`'s greeting block is now a single `greeter$generate_stream(...)` call. `$server()` resolves the base client once and threads it to both session and greeting paths.

State-based backends (Streamlit, Dash, Gradio) now pass `greeting_client_factory=self.greeter.build_client`.

### 4. Cleanup (4bd680e, 3d202b8)

- **Made `greeter` required in Python `mod_server`** — removed dead `else` fallback (all entry points always pass a greeter). Matches R.
- **Eager client resolution in Python** — split `create_client()` into `resolve_client()` (spec → Chat, no clone) and `create_client()` (pure clone). Client specs resolved at init when non-None (fail-fast on invalid providers); deferred pattern (`client=None`) stays lazy. Dissolves the R/Py asymmetry — both now resolve once and clone cheaply. `_client_spec` renamed to `_base_client`; `base`/`greeting_base` types narrowed to `Chat | None`.

gadenbuie:
@cpsievert In looking at this again with fresh eyes this morning, there was a lot of overlapping of concerns in the `build_greeting_client` method and I wasn't a fan. If you want to take another look, the changes are described in the last comment and are here: https://github.com/posit-dev/querychat/pull/261/changes/3b23601739ff6dddaa240df7b3999723ae5a06ae..ffcc17e1982a13d62b5799b45075987438e40666

pkg-py/src/querychat/_shiny_module.py
```diff
@@ -337,7 +339,7 @@ async def _handle_greeting_requested():
             existing = current_greeting.get()
             if existing is not None:
                 await chat_ui.set_greeting(
-                    shinychat.chat_greeting(existing, dismissible=False)
+                    shinychat.chat_greeting(existing, persistent=True, dismissible=False)  # pyright: ignore[reportArgumentType]
```
cpsievert: Wouldn't this error with an old version of Shiny Chat? I say we just require the latest and use the new option.

pkg-py/src/querychat/_querychat_base.py
```diff
@@ -319,10 +327,37 @@ def client(
...
+                sp = QueryChatSystemPrompt(
+                    prompt_template=prompt,
+                    data_sources=self._data_sources,
+                    data_description=self._data_description,
+                    extra_instructions=None,
+                    categorical_threshold=self._categorical_threshold,
+                    data_dicts=self._data_dicts,
+                    include_tables=tables or False,
+                    include_relationships=False,
+                    include_glossary=False,
```
cpsievert, line 352: It isn't immediately obvious to me why you would want to suppress this info the greeting? Also I'm not seeing any public way to opt in to including it?

pkg-py/src/querychat/_shiny_module.py
gadenbuie: No, I fixed that error behavior, so we ignore `dismissible` entirely if `persistent` is provided

pkg-py/src/querychat/_querychat_base.py
gadenbuie: Relationships is suppressed because it relates to joins and AFAICT we can't reliably filter the relationships down to the selected tables, plus the model can't write joins at this step anyway. Similarly, glossary isn't likely to be relevant to the greeting in the same way that it's relevant to the general chat conversation.

I don't think we need a public way to adjust these choices, since the best way to improve the generated greeting is to customize the prompt through `$greeter$prompt`.

pkg-py/src/querychat/_shiny_module.py
cpsievert: Oh, I was talking about an _old_ version, but I can see now that we require the GitHub version, at least on the Python side.

gadenbuie: oh right that makes sense. Yeah, I was assuming we'd require the version that released `chat_greeting()` (v0.4.0+ on the R side) but you're right that we're changing the function signature in a way that this call isn't compatible

gadenbuie: I set this up to use the correct argument for each version, we can easily clean up later as needed when we bump required versions

pkg-py/src/querychat/_querychat_base.py
cpsievert: > plus the model can't write joins at this step anyway

Right, but if we supply column context on multiple tables, I don't think it would be a leap to say the LLM might incorrectly infer relationships, thus leading to "bad" suggestions. Also, if we're pulling in context for the table into the conversation, the LLM may decide it no longer needs to call the the tool to fetch table information -- leading to a situation where it never gets information about the relationships. 

> glossary isn't likely to be relevant to the greeting

Going back to look at how the system prompt behavior works, we actually pull in the glossary into the system prompt (which I think is defendable); so yeah, we shouldn't be also including it here.

gadenbuie: To clarify, the greeting client assembles an entirely new system prompt and does not re-use the main client's system prompt -- it does re-use the system prompt machinery so that users can include the standard user-facing prompt template variables.

From a very high level, I think there are two defensible approaches to the greeting system prompt:

1. Keep the exact same system prompt as we use in the main chat. Ideally, the system prompt is cached and the greeting hits the cached prompt. In this approach, we'd keep tool definitions and we'd just use a user prompt to tell the model not to call tools and to create the greeting. (We also can't adjust the tool definitions without breaking caching either.)

2. Remove tool definitions and optimize the system prompt for the greeting message. In this approach, we don't need to show the model any more information than is required to generate an introductory message that contains decent suggestions.

This PR is a best-effort approach to get minimal but enough information in front of the model, limited to just the tables selected by the user. It also gives devs a way to directly customize the greeting prompt so they can optimize the greeting specifically for their own use cases, but still get generated greetings and even rely on our templating infrastructure.

> > plus the model can't write joins at this step anyway
> 
> Right, but if we supply column context on multiple tables, I don't think it would be a leap to say the LLM might incorrectly infer relationships, thus leading to "bad" suggestions.

The problem is that the relationships aren't structured enough that we can reliably and deterministically subset the included relationships to match the tables whose information we're showing to the model in the greeting's system prompt. I think it's just as likely that showing the relationships for tables that the model doesn't yet know about will lead to hallucinations or "bad" responses.

Primarily, my argument is that whether or not the relationships information is included is over-optimizing at the wrong level. The best way to optimize the performance of the greeting prompt will always be to customize the prompt directly. My take is that we should spend our time getting known high-quality signal into the system prompt and not spend much time worrying about ambiguous information.

cpsievert: > entirely new system prompt

Ah, I had missed that. That also helps me better understand the stance on excluding relationships.

cpsievert (approved):

PR #263, opened 2026-06-25
pkg-py: make py-format

gadenbuie:
For `ruff==0.15.20`


gadenbuie:
never mind, I rolled this into #261

PR #267, opened 2026-07-01
Migrate to shinychat's high-level chat API

cpsievert:
shinychat's high-level chat API is now available in both R and Python, and shinychat is investing further in it — file attachments, cancellation, conversation history, and editable messages, etc. are all moving to live there rather than being something each consuming package reimplements. Migrating querychat onto that API now means we drop our own hand-rolled `ExtendedTask`/cancel/restore/bookmark wiring and instead get these features (and future ones) for free as shinychat adds them, in both languages at once.

The most visible outcome of this migration is history/bookmarking config: instead of two overlapping knobs (`enable_bookmarking` and `bookmark_store`), there's now a single `history` parameter that covers both in-memory chat history and Shiny bookmarking, mirroring shinychat's own API.

### What changes for users

- `enable_bookmarking` is deprecated in favor of `history` (e.g. `history = shinychat::history_options(restore_mode = "bookmark")` / `history=shinychat.types.HistoryOptions(restore_mode="bookmark")`).
- `bookmark_store` is removed from `$app()`/`.app()`; pass `history` instead.
- **Behavior change:** `QueryChat` now defaults to `history = TRUE`/`history=True` (shinychat's own default, browser-localStorage-backed history) instead of no persistence.
- Chat greetings now use shinychat's greeting API — schema-aware, generated on demand, preserved across bookmark/restore.

See the updated `NEWS.md` / `CHANGELOG.md` entries for full detail.

### Test plan

- [ ] R: `make r-check`
- [ ] Python: `make py-check`
- [ ] Manual: bookmark/restore a chat in both R and Python example apps


pyproject.toml
```diff
@@ -22,7 +22,7 @@ maintainers = [
 dependencies = [
     "duckdb",
     "shiny>=1.6.2",
-    "shinychat @ git+https://github.com/posit-dev/shinychat.git",
+    "shinychat @ git+https://github.com/posit-dev/shinychat.git@feat/chat-history",
```
cpsievert:
```suggestion
    "shinychat>=0.6.0",
```

pkg-r/DESCRIPTION
```diff
@@ -59,7 +59,7 @@ Suggests:
 VignetteBuilder:
     knitr
 Remotes:
-    posit-dev/shinychat/pkg-r
+    posit-dev/shinychat/pkg-r@feat/chat-history
```
cpsievert:
```suggestion
    posit-dev/shinychat/pkg-r
```

pkg-r/R/QueryChat.R
```diff
@@ -975,16 +1009,28 @@ QueryChat <- R6::R6Class(
         )
       }
 
+      check_history(history)
+      resolved_history <- history %||% self$history %||% TRUE
+
```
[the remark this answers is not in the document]
cpsievert: This is intentional and mirrors the Python side (`pkg-py/src/querychat/_shiny.py`), where the same deprecated `enable_bookmarking` argument is also a hard no-op + warning, not a soft shim — see `test_shiny.py::test_server_resolves_history_and_warns_on_explicit_enable_bookmarking`, which asserts it 'no longer has any effect.' Keeping R and Python behavior identical here matters more than softening the R-only deprecation window. Leaving as-is.

pkg-r/DESCRIPTION
```diff
@@ -35,7 +35,7 @@ Imports:
     rlang (>= 1.1.0),
     S7,
     shiny,
-    shinychat (>= 0.4.0),
+    shinychat (> 0.4.0),
     utils,
```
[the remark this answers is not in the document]
cpsievert: R and Python shinychat versions are independent (tags are `r/vX.Y.Z` / `py/vX.Y.Z`), so aligning R's minimum to Python's 0.6.0 isn't meaningful here — no such R release exists yet (latest is `r/v0.4.0`). `shinychat (> 0.4.0)` is deliberate: it excludes the released 0.4.0 (which lacks `chat_server()`/`history_options()`/persistent `chat_greeting()`) while accepting the dev version on `main` (`0.4.0.9000`) that the `Remotes:` entry installs, following the same `>` vs `>=` pattern this DESCRIPTION has used historically when depending on unreleased shinychat APIs. Will bump to `>= 0.5.0` (or whatever the R tag ends up being) once shinychat cuts a release with this API. Leaving as-is for now.

gadenbuie (commented):
Two quick things jump out at me from the reading the PR description:

1. > Chat greetings now use shinychat's greeting API — schema-aware, generated on demand, preserved across bookmark/restore.

   I thought we had this already with #261, so I'm guessing this is a bit of Claude missing the nuance in what happened here in the PR. Would you mind clarifying what specific changes we made around greetings?

2. There's a bunch of bookmarking changes in #266 that also touch `enable_bookmarking` and `bookmark_store` and are going to be equally complicated by the new history feature in shinychat. That region of the API is sufficiently complex that I think it's worth us talking through the API design in person to make sure we're coordinating between bookmarking and history in the best way possible.
 
   The core proposal in #266 was to deprecate `enable_bookmarking` in favor of `bookmark_enable` and `bookmark_store`. That's both a prefix hygiene change and a functional change: with the additional cards, `bookmark_enable` would accept both `TRUE` or specific content types that should be included in the bookmark (e.g. `"cards"` or `"chat"`).

   I'm very open to other options for the API, but if there's a chance we'll take #266 (which I do think we should), I think it's worth our time to make sure we're including the cards feature in the bookmarking arguments. The core request there is for it to be possible to restore the cards independently of or together with the chat history.

cpsievert:
> Would you mind clarifying what specific changes we made around greetings?

These changes don't undo the work of #261 -- they just lean into the fact that shinychat can manage the event handling, restoration, etc. 

> I think it's worth us talking through the API design in person

I agree, but I'm hoping we can not let that block a Python release, which I would love to have for SciPy (this Monday). 

If it helps, this PR does the following:

1. Deprecates `enable_bookmark` since we now always register the bookmark save/restore callbacks; and `history=history_options(restore_mode="bookmark")` is the new way to have chat events (response, new chat, etc) trigger a bookmark. 
2. Removes `bookmark_store` from `$app()` since bookmarking chat history is (at least currently) only supported with `bookmark_store="server"`. Also, note that bookmark-based restoration will continue to be the default experience for this method. 

I anticipate if we're going to disagree on anything in the public API surface, it will be in the realm of whether `bookmark_enable` should exist. I would argue that it probably shouldn't because:

1. We can now save and restore the chat (and other state) without bookmarking. 
    * As a result, specifying which portions of app state to save/restore via `bookmark_enable` feels wrong. If we still feel it's important to save and restore only certain portions of state, that should come as a separate parameter -- something like `restore = "cards"`? 
2. It's really just a shorthand for `history_options(restore_mode="bookmark")`

pkg-r/R/QueryChat.R
```diff
@@ -975,16 +1009,28 @@ QueryChat <- R6::R6Class(
...
+      if (!is.null(enable_bookmarking)) {
```
gadenbuie, line 1014: Decision needed: `enable_bookmarking` is deprecation-warned here but its value is then discarded — `resolved_history` is computed purely from `history %||% self$history %||% TRUE`. So `enable_bookmarking = TRUE` now silently resolves to `history = TRUE` (localStorage), not shareable bookmark state. Two coherent options; pick one:

- **Map it (with a warning):** when `enable_bookmarking = TRUE` and `history` was not otherwise set, resolve to `history_options(restore_mode = "bookmark")`. Preserves old behavior through the deprecation window.
- **Error instead of warn:** if we are deliberately not preserving the behavior, a soft deprecation warning is the wrong signal — it implies the code still works as before. An error pointing to `history=` is more honest about the break.

The current middle ground (warn, then change behavior) would be better to avoid.

pkg-py/src/querychat/_shiny.py
```diff
@@ -560,14 +555,29 @@ def title():
         def create_session_client(**kwargs) -> chatlas.Chat:
             return self._create_session_client(base=resolved_client, **kwargs)
 
+        resolved_history: bool | HistoryOptions = (
+            history
+            if history is not None
+            else (self.history if self.history is not None else True)
+        )
+
+        if enable_bookmarking is not None:
```
gadenbuie, line 559: Same decision as the R side (`QueryChat.R:1015`): the deprecation warning fires but `resolved_history` (above) never consults `enable_bookmarking`, so `enable_bookmarking=True` silently degrades to `history=True` rather than bookmark mode. Either map it to `HistoryOptions(restore_mode="bookmark")` with a warning, or raise instead of warn. Whatever we choose, keep R and Python consistent — the R message currently says "for the equivalent behavior" while Python says "no longer has any effect".

pkg-r/R/querychat_module.R
```diff
@@ -116,158 +81,140 @@ mod_server <- function(
...
+      }
+      invisible(NULL)
     }
 
+    # Registered unconditionally: both registrations are no-ops unless the
+    # corresponding persistence layer (Shiny's own bookmarking, or shinychat's
+    # history) is actually active in the host app. In restore_mode = "bookmark",
+    # shinychat's history-save triggers a real Shiny bookmark internally, so both
+    # callbacks fire for the same event -- the snapshot is written twice,
+    # harmlessly, since it's idempotent.
```
gadenbuie, line 208: The state-idempotency claim holds, but this is worth reframing: restore may not be execution-idempotent. In `restore_mode = "bookmark"` both `onRestore` and `history$on_restore` fire, so `apply_state_snapshot()` runs twice, and each pass calls `restore_viz_widgets()` (re-executing every ggsql query) and re-sets every table SQL reactive. Final *state* is identical; the *work* is not free — for several viz widgets over a non-trivial dataset that is real, user-perceived latency on restore. Suggest a sentence here noting restore-time execution can double, so a future reader knows it is a known cost rather than an oversight.

pkg-py/src/querychat/_shiny_module.py
```diff
@@ -367,43 +347,53 @@ def _():
...
+                    last_restored = name
+            if f"querychat_title_{name}" in vals:
+                state.title.set(vals[f"querychat_title_{name}"])
+        if last_restored is not None:
+            _current_table.set(last_restored)
+        if "querychat_viz_widgets" in vals:
+            restored = restore_viz_widgets(executor, vals["querychat_viz_widgets"])
+            viz_widgets[:] = restored
+
+    # Registered unconditionally: both registrations are no-ops unless the
```
gadenbuie, line 380: Same note as the R side: "harmlessly, since it is idempotent" covers the save-side double-write, but on restore both `on_restore` and `history.on_restore` fire in bookmark mode, so `apply_state_snapshot` -> `restore_viz_widgets` re-executes every ggsql query twice and re-sets table SQL twice. State is idempotent; execution cost is not. Worth flagging the doubled restore-time work in the comment.

pkg-r/DESCRIPTION
```diff
@@ -35,7 +35,7 @@ Imports:
     rlang (>= 1.1.0),
     S7,
     shiny,
-    shinychat (>= 0.4.0),
+    shinychat (> 0.4.0),
```
gadenbuie, line 38: Minor housekeeping (not a blocker): `> 0.4.0` does not guarantee the API this PR calls. A dev build (`0.4.0.9000`) satisfies `> 0.4.0` yet lacks `chat_server()` / `history_options()`. Enough has moved upstream in dev shinychat that I think we should bump the dev-version pin here just to be explicit, and leave a NEWS/CHANGELOG note that this floor will need to move to the released shinychat version once that release lands.

pkg-r/NEWS.md
```diff
@@ -33,6 +33,14 @@
 
 * The `$data_source` property has been removed. Use `qc$table("name")$data_source` to read a table's data source, and `qc$add_table(df, "name", replace = TRUE)` to replace it. The `data_source` parameter to `$server()` has also been removed; call `$add_table()` before `$server()` instead. (#195)
 
+* `$app()`/`$app_obj()`'s `bookmark_store` parameter has been removed. Pass `history = shinychat::history_options(restore_mode = "bookmark")` to get the same shareable-bookmark behavior; any other `history` value disables Shiny-level bookmarking for the generated app. `$app()` defaults to `restore_mode = "bookmark"` when no `history` is set anywhere, so existing `$app()` callers keep working without changes.
```
gadenbuie: Worth an explicit line here that the `$app()` default changed from URL-based (`bookmark_store = "url"`) to server-side bookmarking. Local `$app()` users likely will not notice, since shinychat history covers the common case by default — but anyone relying on the old stateless URL behavior would be surprised. Better to call it out than imply pure continuity.

pkg-py/CHANGELOG.md
```diff
@@ -41,6 +41,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * The `data_source` property has been removed. Use `qc.table("name").data_source` to read a table's data source, and `qc.add_table(df, "name", replace=True)` to replace it. The `data_source` parameter to `server()` (Shiny) has also been removed; call `add_table()` before `server()` instead. (#195)
 
+* `.app()`'s `bookmark_store` parameter has been removed. Pass `history=shinychat.types.HistoryOptions(restore_mode="bookmark")` to get the same shareable-bookmark behavior; any other `history` value disables Shiny-level bookmarking for the generated app. `.app()` defaults to `restore_mode="bookmark"` when no `history` is set anywhere, so existing `.app()` callers keep working without changes.
```
gadenbuie: Same as the R NEWS note: worth spelling out that `.app()` shifted from `bookmark_store="url"` to server-side storage by default. Likely unnoticed by local `.app()` users given shinychat history defaults, but a one-line heads-up avoids surprises for anyone who depended on the old URL behavior.

pkg-py/tests/test_shiny_module.py
```diff
@@ -46,3 +46,187 @@ def test_mod_ui_allow_attachments_can_be_overridden():
...
+    inner_fn = _unwrap_module_server(mod_server)
+
+    fake_input = MagicMock()
+    fake_input.__getitem__ = MagicMock(return_value=MagicMock())
+    fake_session = MagicMock()
+    fake_session.is_stub_session.return_value = False
+
+    with (
+        patch(
+            "querychat._shiny_module.shinychat.Chat", side_effect=fake_chat_constructor
```
gadenbuie, line 99: These tests fake `shinychat.Chat` / `chat_server` / `history.on_save` / `enable_bookmarking(bookmark_on=None)`, so an API signature drift upstream would not turn them red — the playwright e2e tests plus the manual bookmark/restore step are the real safety net. Suggestion, not a requirement: it would be good to exercise the real shinychat contract in at least one thin test, and to make sure the e2e + manual steps in the PR test plan run before merge.

pkg-py/tests/playwright/test_11_viz_footer.py
```diff
@@ -9,21 +9,52 @@
...
+    while time.monotonic() < deadline:
+        box = locator.bounding_box()
+        if box is not None and box == last_box:
+            stable_count += 1
+            if stable_count >= quiet_checks:
+                return
+        else:
+            stable_count = 0
+        last_box = box
+        time.sleep(interval_s)
```
gadenbuie, line 67: Two minor points: (1) if the box never stabilizes within `max_wait_s`, this returns with no signal, so a genuine layout regression could masquerade as "settled" — consider failing or warning on deadline exhaustion. (2) dict-equality on `bounding_box()` floats is jitter-sensitive; `expect.poll` would be more idiomatic than the hand-rolled `time.sleep` loop.

pkg-r/R/querychat_module.R
```diff
@@ -116,158 +81,140 @@ mod_server <- function(
...
-          !!!user_input_parts,
-          stream = "content",
-          controller = controller
-        )
+    # Registered unconditionally so the chat client's own state (and greeting)
+    # round-trip through Shiny bookmarks whenever the host app has bookmarking
+    # enabled -- independent of `history`. Only the save/restore hooks are
+    # wanted here; the automatic bookmark triggers are turned off so `history`
+    # (or the host app) remains the sole source of *when* to bookmark.
+    shinychat::chat_restore(
```
gadenbuie: Since we maintain shinychat, this is a question we can answer ourselves rather than defer: is registering both `chat_server()` (above) and `chat_restore()` here against the same `"chat"` id — with auto-triggers off — a supported composition, or is it leaning on incidental behavior? Worth confirming against the shinychat internals before merge.

gadenbuie (approved):
The migration looks great and I agree, we don't need to block a Python release on fully figuring out the bookmarking/history API issues I called out. In fact, I'm confident that we can ship this and then I can work out how to best roll the dynamic cards feature into either history or bookmarking.

I ran the PR by Claude for a review, which flagged a few minor things below. Feel free to resolve and merge!

pkg-r/R/QueryChat.R
cpsievert: Went with mapping: `enable_bookmarking = TRUE` now resolves to `history_options(restore_mode = "bookmark")` when `history` isn't otherwise set, so it keeps its old behavior through the deprecation window. 7688df61

pkg-py/src/querychat/_shiny.py
cpsievert: Same fix on the Python side, plus aligned the two deprecation messages (R already said "for the equivalent behavior"; this one said "no longer has any effect" — both now say the former). 5a946019

pkg-r/R/querychat_module.R
cpsievert: Checked shinychat's actual implementation before touching this: `chat_history.R:718-722` explicitly skips firing the `on_restore` callback when `restore_mode == "bookmark"` (`if (!identical(restore_mode, "bookmark")) { restore_after_first_flush(target$values) }`). So `apply_state_snapshot()`/`restore_viz_widgets()` only runs once, via `shiny::onRestore`, not twice via `history$on_restore` too — the premise doesn't hold, so no change here.

pkg-py/src/querychat/_shiny_module.py
cpsievert: Same check on the Python side — `_history.py`'s `on_restore()` docstring and implementation confirm the identical guard (doesn't fire when `restore_mode="bookmark"`). No change here, same reasoning as the R comment.

pkg-r/DESCRIPTION
cpsievert: Looked into pinning `Remotes:` to a commit, but tested `remotes`'s actual resolution logic first (`remotes::dev_package_deps()`): it diffs the installed package's SHA against the live SHA of whatever ref `Remotes:` points to, not the `Version:` string, and defaults to auto-upgrading on non-interactive installs (interactive sessions get prompted with that diff). So a stale local shinychat gets caught and updated on a normal `install_deps()`/`pak` run regardless of whether we pin — pinning would mostly add manual upkeep (remembering to bump the SHA) without closing a real gap. Left `DESCRIPTION` as it was.

pkg-r/NEWS.md
cpsievert: Added the storage-mechanism callout. 570392bc

pkg-py/CHANGELOG.md
cpsievert: Same callout on the Python side. 570392bc

pkg-py/tests/test_shiny_module.py
cpsievert: Added a thin test that constructs the real `shinychat.Chat` and calls the real `.history.on_save`/`.on_restore`/`.enable_bookmarking()` with the same arguments `mod_server()` uses, so a signature drift in that surface fails loudly here too. 449a15eb

pkg-py/tests/playwright/test_11_viz_footer.py
cpsievert: Addressed both: now `pytest.fail()`s with the last bounding box on deadline exhaustion instead of returning silently, and compares boxes within a small pixel tolerance instead of exact equality. One deviation: playwright-python 1.58 (what's pinned here) doesn't expose `expect.poll()` for a custom condition like this, so the hand-rolled polling loop stays, just made stricter. cb237440

pkg-r/R/querychat_module.R
cpsievert: Confirmed it's leaning on incidental behavior, not a supported composition — shinychat's own `chat_restore()` docs call it and `chat_enable_history()` mutually exclusive, and in bookmark mode both really do independently restore the client's turns (`chat_restore()`'s `onRestore` fires unconditionally; `chat_enable_history()`'s turn-restore is gated on an active bookmark context, not on `restore_mode`). Not a correctness bug in practice — `chat_enable_history()`'s restore always clears before re-rendering, so it's last-write-wins — but it does mean doubled per-turn storage on every save and a restore-time flicker. Fixed by skipping `chat_restore()`/`enable_bookmarking()` whenever `history`'s `restore_mode` is already "bookmark". 5806125c (R), 827e322a (Python)

pkg-r/DESCRIPTION
gadenbuie: Oh the suggestion here wasn't about pinning to SHA but instead about bumping the dev version of shinychat on the shinychat side, e.g. `0.4.0.9001`, as a signal that we've introduced a coordinated API change within the dev cycles of both packages.

cpsievert:
E2E test failures were introduced by a new pyarrow 25.0.0 release. Will fix that separately

PR #274, opened 2026-08-21
feat: frame visualization tool results with Shiny Chat

cpsievert:
## Why this matters
Visualization tool results now use Shiny Chat's native framed presentation, so QueryChat no longer reaches into its condensed tool-display structure to create the surrounding frame.

Builds on [posit-dev/shinychat#331](https://github.com/posit-dev/shinychat/pull/331) (merged; Python and R dependencies now track shinychat main).

## What changes
- Request `presentation="framed"` for visualization results in Python and R.
- Remove QueryChat's Shiny Chat frame-selector overrides while keeping visualization-specific sizing and footer styling.
- Refresh served CSS copies and update focused tests.
- Update viz Playwright suites for the framed DOM (`.shiny-tool-card` inside `.shiny-chat-tool-group`, not `.shiny-tool-result`), and keep the framed viz header/footer undimmed and compact against shinychat main's framed-group CSS.

## Verification
- `uv run pytest pkg-py/tests/test_viz_footer.py -q`
- `uv run pytest pkg-py/tests/playwright/test_10_viz_inline.py pkg-py/tests/playwright/test_11_viz_footer.py pkg-py/tests/playwright/test_12_viz_bookmark.py` — 22 passed
- `Rscript -e ...test-viz-tool.R`

Stacked on #279 (which now carries the viz Playwright DOM-selector updates for shinychat main; this PR keeps only the framed-presentation feature and its tests).

PR #279, opened 2026-08-26
fix: run Shiny E2E apps as subprocesses (demonstrate shinychat main issue)

cpsievert:
The first commit pins shinychat to git main to demonstrate that shinychat main alone (specifically posit-dev/shinychat#280) breaks the Shiny sidebar E2E tests: the in-process threaded uvicorn fixtures share Shiny's process-global, loop-bound reactive lock across apps, and the second app's session dies on lock contention.

The second commit rewrites the Shiny Playwright fixtures to use shiny.pytest.create_app_fixture, which runs each app as a subprocess (matching Shiny's own test suite and our Streamlit fixtures), eliminating the shared-lock hazard.

Expected: E2E fails on the first commit (sidebar title tests), passes after the second.

Later commits adapt the viz Playwright suites (test_10–12) to shinychat main's tool-result DOM (routed results render as `.shiny-tool-card` inside `.shiny-chat-tool-group`; the `.shiny-tool-result` wrapper is now fallback-only) — moved here from #274 — and fix test_13 to use `ShinyAppProc.url`.

### DIFF
--- a/pkg-py/tests/playwright/conftest.py
+++ b/pkg-py/tests/playwright/conftest.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import pytest
+from shiny.pytest import create_app_fixture
 
 # Configure logging for test debugging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


cpsievert:
Update: the pin-only commit (bf87e71e) hit the 30-minute E2E timeout as expected — failed tests included both sidebar title tests (the reactive-lock race) and the test_10 viz inline tests (main's viz tests expect shinychat 0.6.x tool-result DOM; those test updates live in #274).

The fixture-rewrite commit (89caac35) should fix the sidebar title failures. The viz inline failures are expected to remain here since this PR doesn't include #274's test updates.

cpsievert:
Follow-up: the cancel-test failures in the post-fix runs were not caused by the fixture change. The `@main` pin moved from dev38 (`a17dbf3`) to dev40 (`909ed8c`) between CI runs, and dev40 includes posit-dev/shinychat#350, which replaced the `.shiny-chat-btn-cancel` class with a `data-state` attribute on the send button. Verified locally that dev40 breaks those tests with both the old threaded and new subprocess fixtures. The third commit updates the selectors. Remaining expected red: the test_10/11/12 viz tests, which need #274's updates for shinychat main's tool-result DOM.

PR #285, opened 2026-09-03
fix(r): schema-qualify CTE table refs to avoid circular references

cpsievert:
## Problem

Four `TblSqlSource` tests fail with a "circular reference to CTE" error on DuckDB 1.3.0+ (and SQLite):

```
Error: Binder Error: Circular reference to CTE "test_table", There are two possible solutions. 
1. use WITH RECURSIVE to use recursive CTEs. 
2. If you want to use the TABLE name "test_table" the same as the CTE name, please explicitly add "SCHEMA" before table name. You can try "main.test_table"
```

This happens when a `TblSqlSource` is created from a **transformed** `tbl_sql` (e.g., `dplyr::filter()`, `dplyr::select()`). The class wraps the transformed tbl's SQL as a CTE named after the table:

```sql
WITH "test_table" AS (
  SELECT test_table.* FROM test_table WHERE ...
)
SELECT * FROM test_table
```

DuckDB 1.3.0+ and SQLite reject this because the CTE name (`"test_table"`) shadows the physical table (`test_table`) referenced in the CTE body, creating a circular reference.

## Fix

Schema-qualify the table references in the CTE body so they resolve to the physical table, not the CTE:

```sql
WITH "test_table" AS (
  SELECT test_table.* FROM main.test_table WHERE ...
)
SELECT * FROM test_table
```

Two helper functions are added to `TblSqlSource.R`:

- `get_current_schema(conn)` — tries `SELECT current_schema()` (DuckDB, PostgreSQL) and falls back to `"main"` for databases that don't support it (SQLite).
- `qualify_cte_table_refs(conn, cte_body)` — uses a regex to prefix `FROM`/`JOIN` table references with the schema name. Skips already-qualified names and subquery `FROM` clauses.

## Verification

All 64 `test-TblSqlSource.R` tests pass (previously 4 failed). Full `devtools::test()` shows no new failures or regressions.

### DIFF
--- a/pkg-r/R/TblSqlSource.R
+++ b/pkg-r/R/TblSqlSource.R
@@ -171,10 +171,15 @@ TblSqlSource <- R6::R6Class(
         return(query)
       }
 
+      cte_body <- qualify_cte_table_refs(
+        private$conn,
+        as.character(private$tbl_cte)
+      )
+
       sprintf(
         "WITH %s AS (\n%s\n)\n%s",
         DBI::dbQuoteIdentifier(private$conn, self$table_name),
-        private$tbl_cte,
+        cte_body,
         query
       )
     },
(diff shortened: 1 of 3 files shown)


cpsievert:
Closing this PR — the 4 TblSqlSource test failures it addresses are **local-only**, caused by an outdated local DuckDB (1.3.0).

CI uses DuckDB 1.5.5, which already fixed this circular CTE reference bug upstream in [DuckDB 1.5.0 (PR #19116)](https://github.com/duckdb/duckdb/releases/tag/v1.5.0), resolving [issue #6389](https://github.com/duckdb/duckdb/issues/6389). The schema-qualification workaround in this PR adds complexity to work around a DuckDB bug that no longer exists in the versions CI and most users run.

If support for older DuckDB (< 1.5.0) becomes a requirement, this approach could be revisited. Thanks to @cpsievert for the investigation.

PR #245, opened 2026-06-03
feat(handoff): create a standalone report from a chat session 

cpsievert:
## What this adds

querychat is great for *exploring* data conversationally — but today that exploration is trapped in the chat conversation. This PR lets a user **turn the work they did during a querychat session into a standalone, runnable, downloadable handoff**: a Quarto dashboard, a Shiny app, a Marimo or Jupyter notebook, or a freeform format they describe in words.

In short: chat your way to the right queries and visualizations, then use the `/handoff` slash command. You'll be prompted with a dialog to choose relevant findings, an output format (e.g., Quarto, Jupyter, etc), a language (e.g., R and Python), and addition instructions for generating the "handoff" document. This standalone document, a README, as well as the data (if it's small enough), will all be bundled into downloaded zip file. 

**Why it's useful**
- **Bridges exploration → deliverable.** Insights found by asking questions become a reproducible asset.
- **Meets people where they work.** Pick the output format (dashboard / app / notebook) and language (R or Python) that fits the team.
- **Self-contained.** Small datasets are bundled as CSV so the handoff runs as-is; larger / DB-backed sources get a clearly-marked data-setup TODO.
- **Iterative.** Revise with AI and download a zip (source + README + data).

## See it in action

Upon requesting a new handoff:

01-modal

After generating the handoff:

02-panel

## How it works (user's-eye view)
1. Open the **Prepare Handoff** modal — type `/handoff`, or the assistant offers it when it senses you're ready.
2. The modal shows a **gallery** of the session's queries & charts, an **output-format** picker, a **language** picker, and a **directions** box. An LLM recommendation pre-selects sensible defaults while it opens.
3. Hit **Generate** — a side panel slides open and the handoff source **streams** into a read-only editor; a **pill** is added to the chat linking back to it.
4. From the panel: **revise** with AI and **download** the zip.
5. State survives **bookmarking**, so a restored session re-shows its handoffs.

## Approach & architecture (for reviewers)
Built around a few deliberate constraints:

- **A hard reactivity boundary.** `HandoffOrchestrator` is plain async business logic — reads no `reactive.Value`, defines no effects, never touches `input.*`. *All* reactivity lives in `handoff_server()`. That's what lets every flow be unit-tested with plain fakes.
- **One owner per concern.**
  - `HandoffView` — the only place server→client output happens (custom messages, modal, chat pill).
  - `HandoffChat` — the only place chatlas is touched; deep-copies the live chat before each call, so generate/revise/recommend never pollute the visible conversation.
[...]

PR #290, opened 2026-09-08
fix(py): constrain sqlglot<30.18.0 to work around Drop SQL regression

cpsievert:
## Problem

CI's Python test jobs started failing with 48 ibis/duckdb-related test failures/errors, all with:

```
_duckdb.ParserException: Parser Error: syntax error at end of input
```

(e.g. [this run](https://github.com/posit-dev/querychat/actions/runs/34255143465) on #245). These failures are **orthogonal to the changes in those PRs** — they occur in ibis test fixtures (`conn.create_table(...)`) that no branch code touches.

## Root cause

Diffing the installed package set between the last passing run on `main` (2026-09-01) and the failing runs shows the relevant change is **sqlglot 30.17.0 → 30.18.0** (duckdb 1.5.5 and ibis-framework 12.0.0 are identical in both).

sqlglot 30.18.0 regressed SQL generation for `Drop` expressions: the object name is omitted entirely —

```python
import sqlglot.expressions as sge
sge.Drop(kind="VIEW", this="myview", exists=True).sql("duckdb")
# sqlglot 30.17.0: 'DROP VIEW IF EXISTS "myview"'
# sqlglot 30.18.0: 'DROP VIEW IF EXISTS'   <- name missing
```

ibis's duckdb backend issues exactly such a `DROP VIEW IF EXISTS` in `create_table()` when populating a table from an in-memory object ([ibis/backends/duckdb/\_\_init\_\_.py:239](https://github.com/ibis-project/ibis/blob/12.0.0/ibis/backends/duckdb/__init__.py#L239)), and duckdb rejects the truncated statement with the parser error above.

## Fix

Constrain `sqlglot<30.18.0` in the `ibis` extra and the `dev` dependency group, with a comment to remove the pin once the regression is fixed upstream (tobymao/sqlglot).

## Verification

- Reproduced the failure locally with duckdb 1.5.5 + ibis-framework 12.0.0 + sqlglot 30.18.0; confirmed it passes with sqlglot 30.17.0.
- With this constraint applied, `uv run pytest pkg-py/tests --ignore=pkg-py/tests/playwright` passes: **639 passed** (previously 12 failed + 36 errors).

This should also unblock #245 and #286, whose Python test failures have this same cause.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,9 +47,11 @@ polars = ["polars", "pyarrow<25.0.0"] # duckdb requires pyarrow for polars DataF
 ibis = [
     "ibis-framework>=9.0.0",
     "pandas", # required for ibis .execute() to return DataFrames
-    # sqlglot 30.18.0 regressed Drop SQL generation (omits the object name,
-    # e.g. `DROP VIEW IF EXISTS` with no name), breaking ibis's duckdb
-    # create_table(). Remove once fixed upstream (tobymao/sqlglot).
+    # sqlglot 30.18.0 deliberately renamed Drop.this -> Drop.tables
+    # (tobymao/sqlglot#8229), so ibis's Drop(this=...) calls emit e.g.
+    # `DROP VIEW IF EXISTS` with no name, breaking ibis's duckdb
+    # create_table(). Remove once fixed upstream:
+    # https://github.com/ibis-project/ibis/issues/12101
     "sqlglot<30.18.0",
 ]
 pins = ["pins>=0.9.1"]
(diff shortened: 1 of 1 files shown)


cpsievert:
Upstream references for removing this pin:

- Tracking issue: ibis-project/ibis#12101 — `drop_table` / cache cleanup / `create_table` emit empty `DROP ... IF EXISTS` with sqlglot 30.18 (same root cause as the failures this PR works around; ibis's `sqlglot>=23.4` has no upper bound)
- Root cause: deliberate breaking change in sqlglot 30.18.0 renaming `Drop.this` → `Drop.tables` (tobymao/sqlglot#8229)
- Proper fix in flight: ibis-project/ibis#12104 (adapt `Drop` usage across all backends + bump sqlglot to 30.18.0) — open, under review

Once ibis merges #12104 and releases, this pin can be removed in favor of requiring the fixed ibis version.

Issue #300, opened 2026-09-12
server(data_source = ) is unsafe for concurrent sessions (R and Python)


cpsievert:
## Summary

Some deployments can only create a table's data source *inside* the Shiny server function — for example, on Posit Connect, a per-user database connection (e.g. Snowflake via `PositAuthenticator`) depends on that user's session-scoped OAuth credentials, which aren't available until the server function runs. Both R and Python have a `server(data_source = )` pattern meant to support exactly this: register a table lazily, once the session starts, instead of at app-startup time.

That pattern is currently broken in both languages — not just missing in Python, but genuinely unsafe for the concurrent-session case it exists for.

## Example

```r
qc <- QueryChat$new(NULL, table_name = "orders")  # no data source yet

server <- function(input, output, session) {
  conn <- DBI::dbConnect(..., token = session_user_token(session))  # per-user connection
  qc$server(data_source = conn)
}
```

- **User A** opens the app. Session A's `server()` runs and registers `conn_A` under `"orders"`. Works.
- **User B** opens the app — a second browser tab, or simply a second person hitting the same running process. Session B's `server()` runs and registers `conn_B` under the same name.

Depending on the language, and how far you dig, different things go wrong here:

1. **It can outright fail for the second session.** Registering a table checks a guard meant to stop *you* from reconfiguring tables once a conversation is already underway — but that guard trips on any second session, not just a manual reconfiguration. In R, this means session B's `qc$server(data_source = conn_B)` call aborts with "Cannot add tables after server initialization." (Python's `server()` sidesteps this today only because it dropped the `data_source` parameter entirely — see History below.)

2. **Even once that's bypassed, it can corrupt session A.** "Registering a table under a name that's already taken" is implemented as "replace the old one," which includes cleaning up the old table's resources — closing a database connection, disposing an engine, etc. If session A is still running when session B registers `conn_B` under `"orders"`, that cleanup closes `conn_A` — session A's *own* connection — out from under it. Session A's next query then fails, even though session A did nothing wrong and has no relationship to session B.

3. **The welcome greeting can also leak between sessions.** When no explicit greeting is provided, one is generated by asking the LLM to describe the registered table(s) — asynchronously, after the server function has already returned. If session B registers its table in the gap between session A's server function returning and session A's greeting actually being generated, session A's greeting can end up describing session B's table instead of its own.

## History / how this was found

[...]

PR #306, opened 2026-09-12
fix(r): let $server(data_source=) survive a second session safely

cpsievert:
## Stacked on #305

Depends on #305 (deferred-construction ergonomics) — review that first. Two notes on top of the original change:

* #305 now fixes the module-ID desync by **removing** `$add_table()`'s single-table id auto-rename entirely (matching Python's set-id-once semantics), rather than pinning the id — so there's no `.id_pinned` guard to carry into `add_or_replace_table()` here; the rename block is simply gone.
* `greeter$tables` is now deduplicated, so per-session registration doesn't append the same table name once per session.

## The problem

Some deployments can only create a table's data source *inside* the Shiny server function — for example, on Posit Connect, a per-user database connection depends on that user's session-scoped OAuth credentials, which don't exist until the server function runs. querychat's R package already supports this "deferred" pattern: construct `QueryChat$new(NULL, table_name = ...)`, then register the real data source per session via `$server(data_source = ...)`.

In practice this only ever worked for the *first* session:

```r
qc <- QueryChat$new(NULL, table_name = "orders")

  conn <- get_per_user_connection(session)  # depends on this session's OAuth token
  qc$server(data_source = conn)
}
```

- A second browser tab, page refresh, or second concurrent user hit `"Cannot add tables after server initialization."` — `$server(data_source=)` calls the public `$add_table()`, which refuses to run once a session has started.
- Even bypassing that guard, `$server(data_source=)` registers under `replace = TRUE`, and replacing a table cleans up the one it replaces — closing a database connection, disposing a cached query executor — even though an earlier, still-running session's own chat may still be querying through it.
- The auto-generated welcome greeting is built lazily/asynchronously from live, mutable state (`private$.data_sources`, `greeter$tables`, `private$.data_description`), so a later session's registration could leak into an earlier session's greeting.

This mirrors the equivalent bug already fixed in Python (posit-dev/querychat#302, #303, #304); see posit-dev/querychat#300 for the original writeup covering both languages.

## The fix

- `$add_table()`'s guard-free core (name validation, staging the table, rebuilding the system prompt/executor cache, greeting inclusion) is extracted into a private `add_or_replace_table()`. `$add_table()` keeps its guard and calls into that core; `$server(data_source=...)` calls the core directly, bypassing the guard for this path only.
- Live sessions are tracked on the instance (`.active_sessions`, incremented per `$server()` call, decremented via `session$onSessionEnded()`), and shared-state decisions key off it:
  - `add_or_replace_table()` gains `cleanup_replaced = TRUE`; `$server()`'s internal call passes `cleanup_replaced = private$.active_sessions == 0`, so a replaced resource (and the cached query executor) is cleaned up exactly when no running session could still be using it. The public `$add_table(replace = TRUE)` is unaffected — that path still has exactly one owner.
  - The `$add_table()`/`$add_tables()`/`$remove_table()` guards block only while a session is actually live, rather than forever after the first `$server()` call — so a resource registered by a session that has since ended is reclaimed when a later session replaces it.
  - Fails safe: if `onSessionEnded` never fires, behavior degrades to over-conservative cleanup skipping, never premature cleanup.
- `$server()` captures `greeting_tables` and `greeting_data_description` snapshots synchronously at call time and threads them (along with the already-per-session `data_sources` snapshot `mod_server()` already receives) through to `greeter$build_client()`, so the lazily-invoked greeting is built from point-in-time snapshots instead of reading live, mutable state whenever the greeting callback actually runs.

Matches the Python side, #308.

## Test plan

- New tests in `test-server_data_source.R`: a second session's `$server(data_source=)` call succeeds without raising; the public `$add_table()` guard is enforced while a session is live; a second session's call does not clean up the first session's data source or cached query executor (while the public `add_table(replace=TRUE)` path still does, locking in unchanged behavior); `$server()` passes `greeting_tables`/`greeting_data_description` snapshots to `mod_server()`, which forwards them (with the per-session `data_sources`) to `greeter$build_client()`; per-session registration does not duplicate `greeter$tables`; mixing config-time `$add_table()` with per-session `$server(data_source=)` behaves as documented (unnamed registration replaces the config-time table (cleaned up when no session is live, left alone while one is), an explicit `table_name` adds alongside it, and the registry is shared across sessions).
- New session-lifecycle tests: a replaced source is cleaned up once the session that registered it has ended; a replaced source survives while *any* session is live; `$add_table()` is allowed once all sessions have ended.
- `air format --check` is clean and the full test suite passes locally (2123 tests), including #305's deferred-construction tests.

## Review follow-ups (Copilot)

- Resources replaced while sessions are live are now retained (`.retired_resources`) and cleaned up when the last live session ends (or `$cleanup()` runs) — closing the executor/connection leak in the `cleanup_replaced = FALSE` path. Matches the Python side (#308).
- `greeter$build_client()` distinguishes an omitted `data_sources`/`data_description` override (live fallback) from an explicit snapshot, which may itself be `NULL` — a session whose snapshot had no inferred description can't pick up a later session's.
- `QueryChat$new(NULL, table_name = ...)` validates the deferred name at construction instead of failing later at `$server()` registration.


cpsievert:
Copilot review feedback addressed in 278adee3:

1. **Unconditional `cleanup_replaced = FALSE`** — now conditional on `private$.server_initialized`: the first `$server(data_source=)` call cleans up what it replaces (no session can be using it yet), restoring the long-standing single-owner behavior; only subsequent sessions skip cleanup. Covered by new tests, and the mixed-API test was updated to match.
2. **`data_description` read live by the greeting factory** — the greeting snapshot now also captures it: `$server()` passes `greeting_data_description` through `mod_server()` to `greeter$build_client()`, which threads it into the greeting's system prompt. Covered by new tests at both the `$server()` and `build_client()` levels.
3. **Test fake not asserting snapshot forwarding** — the fake greeter now records all four arguments and the test asserts `tables`, `data_sources`, and `data_description` are forwarded, with the lazy greeting callback invoked.

Note on the Python side: (1) applies there too and is fixed in #308; (2) does not — Python's `_data_description` is fixed at construction and inferred descriptions live on the `DataSource` objects, which the existing `data_sources` snapshot already covers; (3) was already asserted in Python's `test_shiny_module.py`.

Also: this branch and #305 have been rebased onto `main` now that #301–#304 are merged.

cpsievert:
Copilot review feedback addressed in 9cde4cdb:

1. **Shared underlying DBI connection disconnected on flush** — new `DBISource$get_connection()` plus a `shares_underlying_connection()` guard: when the replaced and replacing sources wrap the same connection (two sessions each passing the same raw `DBIConnection` get their own wrappers), the replacement inherits ownership and the old wrapper is neither cleaned nor retired. Applied to both the `$server(data_source=)` retirement path and `$add_table(replace=TRUE)`. Covered by new tests (same-conn replacement survives flush; different-conn replacement still gets disconnected; `$add_table(replace=TRUE)` keeps a shared connection open).
2. **`$cleanup()` could skip the retired-resource flush on error** — the flush is now registered with `on.exit()`, so it runs even if cleaning the current executor or a source throws. Covered by a new test.
3. **Session counted before setup completed** — `$server()` now increments `.active_sessions` (and arms the `onSessionEnded` flush) only after `mod_server()` returns successfully, so a failed call on a still-live session no longer blocks table mutations or defers replacement cleanup indefinitely. Covered by a new test (failed `$server()` leaves `$add_table()` unblocked).
4. **Compatibility check leaked the staged source** — `check_source_compatibility()` moved inside the cleanup-on-error `tryCatch` in `add_or_replace_table()`, so a failed per-session registration cleans up the staged source's connection. Covered by a new test asserting the connection is disconnected.
5. **`client_factory` contract doc missing `data_description`** — fixed in `QueryChatGreeter`.
6. **Lazy greeting snapshot args in `mod_server()`** — `greeting_tables`/`greeting_data_description` are now `force()`d at module entry, so the deferred greeting callback sees the session's point-in-time snapshot, not a later session's mutation. Covered by a regression test that mutates the promised bindings between setup and greeting time (verified to fail without the fix).

Note: (1) slightly changes `$add_table(replace=TRUE)` semantics beyond this PR's scope — replacing a table with the *same* connection no longer disconnects it (previously it did, breaking the replacement source).

Full `pkg-r` test suite passes.

cpsievert:
Copilot review feedback addressed in 3c6332d9:

1. **Alternating connections leak (C1 → C2 → C1)** — the previous fix only compared the immediately replaced source with the new one, so a retired wrapper from an earlier registration could still share its connection with the now-current source and get disconnected by the flush. `flush_retired_resources()` now skips any retired source that shares its underlying DBI connection with a *current* data source. Covered by a new test (registrations alternating between two connections; the live connection survives the flush, the other retired wrapper is still cleaned up).
2. **Failed retired-resource cleanups dropped permanently** — `flush_retired_resources()` cleared the list up front and suppressed errors, so a transient cleanup failure made the resource unreachable and unretryable. Failed entries are now retained and retried by the next flush (or `$cleanup()`). Covered by a new test with a fail-once cleanup spy.
3. **Failed registration not atomic** — `check_source_compatibility()` now runs before `auto_fill_data_description()` mutates anything, and both `.data_description` fields are snapshotted and restored on any staging error, so a failed per-session registration leaves the existing registration (including its inferred description) fully intact. Covered by a new test using a description-carrying source and a failing prompt build.

All three regression tests were verified to fail when the fixes are reverted. Full `pkg-r` suite passes (2152 pass, 0 fail).

cpsievert:
Closing without merging. #305, which this was stacked on, stands on its own and will be merged separately.

This PR is a faithful port of the Python session-safety work (#302, #303, #304, #308, #309), and that stack is now being replaced rather than extended. The root problem in both languages is that `$server(data_source=)` mutates the app-global `QueryChat` instance from inside a session, which forces all the live-session counting, retired-resource deferral, and greeting snapshots. The redesign makes sessions read-only against the instance: `$server(data_source=)` builds a session-local set of sources, executor, system prompt, and greeting tables, passes that to `mod_server()`, and cleans up what it created in `session$onSessionEnded()`. Porting that to R directly is less work than porting it on top of this PR and deleting most of it again.

Known trade-off: R `main` still has the second-session `"Cannot add tables after server initialization"` error for the deferred pattern until the redesign lands. If that drags on, this PR is the fallback stopgap.

Tests here worth rewriting against the new design: a second session's `$server(data_source=)` succeeds, does not close the first session's source or executor, and the greeting describes the table the session actually registered.

